PR #16, opened 2015-07-19
correction of some typos

Ivo-Balbaert:
Hi Pony creators,

I have worked through the tutorial which is really a nice read, and from what I have seen and understood I like Pony very much, congratulations! Along the way I have collected the typos I saw, most of them are trivial text typos, a few in code sections also: I have corrected them all in this pull request.

I also have a few suggestions which I list here:

1) Section:  Actors - Behaviors
why is amount.min used in this code:
	be eat(amount: U64) =>
    	_hunger_level = _hunger_level - amount.min(_hunger_level)
amount.min has not been introduced yet

Perhaps replace it by:    	_hunger_level = _hunger_level - amount
Which is easier to understand at this level in the tutorial.

2) Section: Control structures - Loops
while count <=  do10 do
  env.out.print(count)
  count = count + 1
end
<-- In this code do has not the keyword colour as have while and end

3) Section: Capabilities – Guarantees 
If an actor has a val variable, no other variable can be used by any actor to write to that object.
Shouldn’t this be:
If an actor has a val variable, no variable can be used by any actor to write to that object.
(leave out other)
By saying no other variable it is suggested that it can be written to with this val variable, if I understand correctly.

4) Section: Aliasing – Ephemeral types
In Pony, every expression has a type. So what's the type of consume a?
	<-- no typo here, but I am pretty sure that the consume keyword has not been talked about until now.
(perhaps a reference to the following section is sufficient)

5) Section: Recovering capabilities – How does this work?
Inside the recover expression, your code only has access to sendable values from the enclosing lexical scope.
	<-- no typo here, but it was not clear to me what sendable means, please explain

Keep up the great work, I will be following Pony with great interest!

Ivo

Signed-off-by: Ivo Balbaert <ivo.balbaert@telenet.be>

### DIFF
--- a/docs/c-ffi/calling-c.md
+++ b/docs/c-ffi/calling-c.md
@@ -61,7 +61,7 @@ Pony classes corresponds directly to pointers to the class in C.
 
 For C pointers to simple types, such as U64, the Pony `Pointer[]` polymorphic type should be used, with a `tag` reference capability. `Pointer[U8] tag` should be used for void*. This can be seen in our SSL_CTX_ctrl example above.
 
-To pass pointers to values to C an ampersand (&) can be used, just like taking an address in C. This is done in the standard library to pass the address of a U64 to an FFI functions that takes a uint64_t* as an out parameter:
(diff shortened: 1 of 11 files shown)


sylvanc:
Thanks! I've deployed your changes, and I'll get to your suggestions as soon as I can.

PR #19, opened 2015-07-25
Editors update: emacs and vim

arialdomartini:
Hi.
I saw that emacs was not present in the list of editors supporting Pony, despite the fact that a [major mode](http://www.vim.org) was already been already published, so I added it.

Also, all the editors had a link to their official websites, but vim. I fixed that.

Cheers!


sylvanc:
Thanks!

PR #24, opened 2015-11-08
Add some examples

pyros2097:
I have added some things, I guess some parts need to be removed as they are todo.


sylvanc:
Thanks!

PR #25, opened 2015-11-09
Document the basic use of delegates in pony

darach:
Add some basic documentation on the use of delegates.

@sylvanc @andymcn Let me know if I got anything wrong and/or need to tweak anything

### DIFF
--- a/docs/types/delegates.md
+++ b/docs/types/delegates.md
@@ -1,0 +1,157 @@
+Like some other object-oriented languages, Pony has __delegates__. That is, some types serve as __mixins__ that other types can use.
+
+# Delegates
+
+Any __trait__ or __interface__ can be used as a _mixin_ by exploiting delegate type declarations.
+
+```pony
+trait Wombat
+  fun box battle_call() : String val => 
+    "Huzzah!"
+
+class SimpleWombat is Wombat
+
+actor Main is Wombat
+  let w : SimpleWombat delegate Wombat = SimpleWombat
+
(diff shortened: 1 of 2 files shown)


sylvanc:
Nice!

andymcn:
Thanks for the docs. However, the line
```
let w : SimpleWombat delegate Wombat = KungFuWombat
```
is an error, since `SimpleWombat` and `KungFuWombat` are both classes and so not subtypes of each other.

Any chance you could check that all your other examples actually compile?

darach:
@andymcn Well spotted. Somehow I missed this copying between ST and the doc. PR to follow to resolve.

andymcn:
Having reread this I have a further change request. It occurs to me that you haven't actually explained what delegation **is**. You've said it's like mixins, which great for people who are used to languages that have mixins, but we'd like to accessible to programmers from as many backgrounds as possible. In particular, some languages use the word delegate to mean slightly different things. (In C# it's effectively a function pointer, seriously, they called that a delegate!)

You say "The actor `Main` delegates its __Wombat__ness to the `SimpleWombat` class" which is a good start (and bonus points for the word wombatness). Could we expand upon that slightly? Something along the lines of "whenever a method in the `Wombat` trait is called on `Main` that call is redirected to the `SimpleWombat` stored in the `w` field" maybe?

darach:
@andymcn Agreed. I'll update later today, opportunity permitting.

PR #26, opened 2015-11-09
Fix typo spotted by @andymcn that krept into the docs

darach:
@andymcn Is it worth adding the final delegate example file to examples in ponyc?


andymcn:
@darach, yes I think putting that in examples would be good.

PR #37, opened 2016-01-12
c-ffi/calling-c: updates, fix &

rurban:
replace & by addressof
add EGLEvent type decl to the setEvent example.
add some better formatting


andymcn:
Thanks for the update.

PR #38, opened 2016-01-12
platform as use condition, editor already

rurban:
Various cleanups: 

what-you-need-to-start already contains the empty
editor-plugins section.

split use-statement into the 3-4 wanted pages: package-system,
use-statement, builtin and the new use conditionals as Platform dependant code

add the missing compiler-clargs page.


andymcn:
Thanks for the edits. The Pony devs are all terrible at getting round to writing docs (we want to write code), so feel free to write lots for the tutorial if you want to :)

I think it's a mistake to list the command line arguments like that. It's just going to get out of date really quickly and all the information is available simply by running `ponyc --help`.

Instead it would be better to give details of a few of the common commands that newbies are likely to need and say use `--help` to see the rest. Potentially useful ones to explain include `--debug`, `--path`, `--output` and `--docs`. It's probably also worth mentioning there are command line parameters automatically handled on programs built with ponyc to tweak runtime performance (also detailed with `ponyc --help`).

andymcn:
Also you mean "dependent", not "dependant". A dependant is a person who depends on some one else (usually financially), such as a child.

rurban:
Agree with the overlong command line list. Will try to come up with something.

rurban:
So just the generics are missing now.

andymcn:
In the command line arguments section you say that `--docs` "dumps the list of classes, actors and types with some warnings to stdout." This is worrying me slightly. Are you saying that the docs are dumped to stdout or that when you generate the docs you're getting warnings printed to stdout? Either way that shouldn't be happening, please open an issue for this with an example output.

I'm going to merge this now anyway, but I might tweak your wording once this is sorted out.

Issue #34, opened 2015-12-07
Tutorial doesn't mention ISize and USize types


jkleiser:

jkleiser:
And sandbox.ponylang.org doesn't understand ISize (and probably not USize).

jonas-l:
I think both tutorial and sandbox are updated to match latest released version. Which is now [0.2.1][releases] and there are [quite a few features undocumented in tutorial and unavailable in sandbox][enhancements-after-0-2-1].

[releases]: https://github.com/CausalityLtd/ponyc/releases
[enhancements-after-0-2-1]: https://github.com/CausalityLtd/ponyc/issues?q=is%3Aissue+is%3Aclosed+closed%3A%3E2015-10-06+label%3Aenhancement

jkleiser:
From the description under "Built-in primitive types" here
http://tutorial.ponylang.org/types/primitives.html
it is easy to understand what types like I8 and U8 mean, but ISize, ILong, USize and ULong are not that obvious to me.
On of the Guiding Principles of the tutorial is this:

> Fully defined semantics. The semantics of all language features must be available in the standard language docs. It is not acceptable to leave behaviour undefined or "implementation dependent".

A few words should be all it takes to explain ISize, ILong, USize and ULong.

SeanTAllen:
Those are actually Guiding principles for Pony not specifically the tutorial. 

Are you able to issue a PR for ISize, ILong etc?

jkleiser:
No, I'm sorry.

jkleiser:
I just issued a PR for this. Thanks to Joe Eli McIlvain. ;-)

PR #39, opened 2016-01-12
Add USize, ISize issue #34

rurban:


andymcn:
Can you also add ULong and ILong please.

rurban:
Sure. 

Are there no type aliases for Int, UInt or Float/Double btw? I find them very convenient in my type system.

andymcn:
No we don't have aliases for those. We're very much against numeric types whose size is not explicit as this leads to loads of bugs when code is ported to platforms with different sizes.

Originally all the Pony numeric types had explicit sizes, but we added the size and long ones because they're needed for writing cross platform code. However, we recommend that they're only used when absolutely needed and we won't be adding any more variable size types unless we have to.

Note that the size and long types are not just aliases for some other integer types, they are actually different types, and explicit conversions to/from the fixed size types are needed. This approach is needed to avoid the problems C has with variable size types.

For Int and UInt use the size you actually want.

F32 is exactly C float and F64 is exactly C double, so there's no need for aliases anyway.

rurban:
I thought not of Int or Float type-alias for an API, you are right. This is wrong. I made the same mistake, bad C habits, perl6 just fixed that.
I thought just for convenience in the matcher, e.g.

    match x
    | let i :Int => "Int"

instead of listing all Int types there.

andymcn:
Sorry, I misunderstood what you were asking there. We do have union types defined that may be helpful to you. The following are defined in `builtin/arithmetic.pony`:
```pony
type Number is (Signed | Unsigned | Float)
type Signed is (I8 | I16 | I32 | I64 | I128 | ILong | ISize)
type Unsigned is (U8 | U16 | U32 | U64 | U128 | ULong | USize)
type Float is (F32 | F64)
```

So you can do:
```pony
match x
| let i: Signed => ...
end
```

PR #40, opened 2016-01-12
Add some runtime pages

rurban:
from some tutorials and papers

only the run-time memory section is pretty empty, esp. regarding managing external memory (as c lib, or via ffi). typical safeties apply, but samples are needed.


rurban:
And people are right. There really need to be a few more cross-links. Will add them...

PR #53, opened 2016-03-13
Document 'with' blocks

georgemarrows:
Based on @sylvanc's explanation here: https://groups.io/g/pony+user/message/106 . I couldn't find them covered in the tutorial; it seemed they belonged best with the rest of exception handling.

### DIFF
--- a/docs/expressions/exceptions.md
+++ b/docs/expressions/exceptions.md
@@ -108,6 +108,14 @@ else
 end
 ```
 
+Multiple objects can be set up for disposal:
+
+```pony
+with obj = SomeObjectThatNeedsDisposing(), other = SomeOtherDisposableObject() do
+  // use obj
+end
+```
+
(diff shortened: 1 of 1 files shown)


SeanTAllen:
thanks @georgemarrows!

docs/expressions/exceptions.md
```diff
@@ -78,6 +78,46 @@ __Do I have to have an else error handler to have a then block?__ No. You can ha
...
+end
+```
+
+`obj.dispose()` will be called whether the code inside the `with` block completes successfully or raises an error. To take part in a `with` expression, the object that needs resource clean-up must therefore provide a `dispose()` method:
+
+```pony
+class SomeObjectThatNeedsDisposing
+  // constructor, other functions
+
+  fun dispose() =>
```
jemc: This may or may not matter but as a small comment - it seems in most (if not all?) cases, cleaning up resources would require more than just read-only access to the object.

To avoid confusing readers, perhaps it would be most idiomatic for this to be `fun ref dispose`?

PR #56, opened 2016-03-23
Document lambda syntax for closures

doublec:
This is based on [my blog post](https://bluishcoder.co.nz/2016/03/15/closures-in-pony.html). I wasn't sure how much detail to go into so welcome feedback on structuring it.


SeanTAllen:
Definitely an improvement. Thanks @doublec.

jemc:
@doublec Not to be overly pedantic here, but I would be careful with terminology.

Capturing in a lambda does not really make the lambda a *closure*.  Yes, if you capture a mutable reference, you can mutate it from within the lambda, but a captured reference cannot change what the outer reference *points to*, which I think is a key part of being a closure.

For example, if you capture an `var x: U8 = 0` reference within a lambda, you cannot use that lambda to make `x = 1` in the outer context - you only end up changing the field within the lambda's object.

SeanTAllen:
@jemc that is a very good point. I think an update makes sense. @doublec you game?

doublec:
> For example, if you capture an var x: U8 = 0 reference within a lambda, you cannot use that lambda to make x = 1 in the outer context - you'd only end up changing the field within the lambda's object (if this is even possible - I haven't checked).

That is correct - it doesn't change the original reference but does change the field within the lambda's object as they are expanded in the object as 'var' not 'let' fields. 

My understanding of the term closure is that it doesn't require being able to mutate the original variables on the stack - it can just capture the value. See [Lexical Environment](https://en.wikipedia.org/wiki/Closure_%28computer_programming%29#Lexical_environment) in the wikipedia entry for example. 

> On the other hand, many functional languages, such as ML, bind variables directly to values. In this case, since there is no way to change the value of the variable once it is bound, there is no need to share the state between closures—they just use the same values. This is often called capturing the variable "by value". Java's local and anonymous classes also fall into this category—they require captured local variables to be final, which also means there is no need to share state.

I'm happy to change what I wrote though if it's preferred not to use 'closure' - what's a better term to differentiate between a function that doesn't capture and one that does?

jemc:
@doublec - Fair enough, thanks for broadening my understanding of what - I'll take wikipedia as a good enough source to tell me my understanding is wrong :smile:.

PR #65, opened 2016-03-31
Update default capability syntax

Praetonus:
The syntax for types default capabilities changed recently but the tutorial still uses the old version.


jemc:
Thanks!

PR #87, opened 2016-04-07
Add C FFI callbacks documentation

### LINKED ISSUE #83: Add a section for callbacks in FFI documents
There are some subtleties to setting up callbacks with C libraries/frameworks that I think would be useful to cover in a tutorial section.

* receiver is first argument
* how to pass function (`address of [object].[function]`)

aturley:
This documentation explains some of the subtle issues around passing
Pony functions to C libraries that will use them as callback
functions. It uses the SQLite `sqlite3_exec` function as an example.

Fixes #83

### DIFF
--- a/docs/c-ffi/callbacks.md
+++ b/docs/c-ffi/callbacks.md
@@ -1,0 +1,113 @@
+# Callbacks
+
+Some C APIs let the programmer specify functions that should be called
+to do pieces of work. For example, the SQLite API has a function
+called `sqlite3_exec` that executes an SQL statement and calls a
+function given by the programmer on each row returned by that
+statement. The functions that are supplied by the programmer are known
+as "callback functions". Pony functions can be passed as callback
+functions, but there are a few conditions that must be met.
+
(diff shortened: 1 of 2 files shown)


docs/c-ffi/callbacks.md
```diff
@@ -0,0 +1,113 @@
...
+`xCallback` becomes the receiver in the call to the Pony function, all
+subsequent arguments are passed along to the Pony callback function in
+order. The parameters of `callback` are arranged to reflect this.
+
+If the order of the arguments to the callback function are not such
+that the receiver object will be passed as the first argument, the
+programmer will need to write a "shim" function in C that accepts
+arguments in the order provided by the API and rearranges them to make
+a call to the Pony function with arguments in the order that Pony
+expects.
```
jemc: Note that we are planning to provide a pure pony way to do this with no shim: https://github.com/ponylang/ponyc/issues/701.  Maybe we should mention here that it is on the roadmap (or maybe we shouldn't - I don't know).

PR #89, opened 2016-04-08
Finalisers documentation

Praetonus:
Finalisers were missing from the documentation. This fixes the problem.


jemc:
Perfect. Thanks!

PR #103, opened 2016-04-17
unescaped pipe character

MrMormon:
One wasn't escaped in the first place, but I think we need `&#124;` because the [online tutorial](http://tutorial.ponylang.org/appendices/symbol-lookup-cheatsheet.html) doesn't understand `\|`.

### DIFF
--- a/appendices/symbol-lookup-cheatsheet.md
+++ b/appendices/symbol-lookup-cheatsheet.md
@@ -34,7 +34,7 @@ Here is a more elaborate explanation of Pony's use of special characters:
 |    | (2) middle of line: method call
 | () | (1) parentheses, for function or behavior parameters 
 |    | (2) making a tuple (values separated by ,)
-|    | (3) making an enumeration (values separated by |)
+|    | (3) making an enumeration (values separated by &#124;)
 | [  | (1) start of line: start of an array literal
 |    | (2) middle of line: generic formal parameters
 | [ ]  | (1) to indicate a generic type, for example Range[U64]
@@ -57,7 +57,7 @@ Here is a more elaborate explanation of Pony's use of special characters:
 |    | (2) to ignore a tuple item in a pattern match
 | ~  | partial application
 | ^  | an ephemeral type
-| \| | (1) separates the types in an enumeration (the value can be any of these types)
+| &#124; | (1) separates the types in an enumeration (the value can be any of these types)
 |    | (2) starts a branch in a match
 | &  | (1) separates the types in a complex type (the value is of all of these types)
 |    | (2) intersection


jemc:
To me it seems more appropriate to surround all of these special characters in backticks instead, so that they will show up as `|` instead of |

Also, then the tutorial is still readable in markdown format in a text editor.

Thoughts?

MrMormon:
I didn't notice until you made me look, but using backticks seems to be the style of the rest of the tutorial. Should I close this and try again with the whole page/section?

jemc:
You can close if you like - or simply change the commit and commit message and title of the PR - whichever is easiest.

SeanTAllen:
The bug that is causing this issue is supposed to fixed in the not so distant future. I'd rather not make any changes such as surrounding everything in backticks that are going to be a large amount of work to change unless the idea is that it should remain that way.

I've known about this issue and left as is, hoping for there to be a fix by the end of April and if there isn't, deciding what to do then.

MrMormon:
@SeanTAllen There's still a pipe on the cheatsheat page that isn't escaped like the other one if I may fix it. Also, the rest of the tutorial uses backticks all the time, so is the issue that it shouldn't be necessary over GitHub's '\' recommendation for single characters?

SeanTAllen:
Definitely should fix the unescaped one.

I don't understand the other bit.  I'm not sure what you mean by "the rest of the tutorial uses backticks all the time, ...". Uses backticks all the time when?

That entire table should have | at the end as well. I thought it did. I requested that before merging. Not sure what happened there. I'm going to fix that now.

I'm ok with using backticks around all the symbol table entries as long as everyone is agreed that we are doing it because we want to not to work around a temporary formatting problem on the githook side.

SeanTAllen:
@MrMormon i just found the unescaped | you referenced. Yeah, that makes sense as a backtick in that scenario. Thoughts on if the symbols in the first column should have backticks?

SeanTAllen:
So looking at this there are also lines like:

```
|    | (2) making a tuple (values separated by ,) |
```

Should the , be in backticks? I'm trying to think of how far to go with that on the 2nd column. I'm thinking yes , gets backticks in that case.

MrMormon:
@SeanTAllen > I don't understand the other bit. I'm not sure what you mean by "the rest of the tutorial uses backticks all the time, ...". Uses backticks all the time when?

Not for single characters necessarily, but \`language element or code example\` is used on nearly every page of the tutorial.

I'm new to Git and don't know if I can change my PR, so maybe you'd please be willing to fix this issue in your PR instead? Sorry.

> Thoughts on if the symbols in the first column should have backticks?

@jemc or someone else's opinion would matter more than mine, but that would make the font style uniform with a \`\\|\` fix.

SeanTAllen:
@MrMormon i can do that. if you'd like help sometime learning some git basics, let me know. i'm far from an expert but can help with some basics. if you are looking to learn in general https://git-scm.com/book/en/v2 is an excellent online resource.

jemc:
I think it makes sense to use backticks around all code snippets - even when those code snippets are just one character.  I think it makes it more readable in general and distinguishes code snippets from punctuation and symbols in ordinary text.

> I'm ok with using backticks around all the symbol table entries as long as everyone is agreed that we are doing it because we want to not to work around a temporary formatting problem on the githook side.

Right, my preference is not about working around the escaping bug - but not needing to escape is a nice side effect.

MrMormon:
I think we're agreed that backticks are the preferred fix for both pipes, but a PR would do the same for other special characters on the page too.

Issue #99, opened 2016-04-16
Please add definition of valid type/interface/trait/primitive/class/actor names


MrMormon:
- ascii letters
- can't start with number
- (double) underscores
- primes at end

SeanTAllen:
This is already covered in the tutorial.

http://tutorial.ponylang.org/types/classes.html

If there is something you feel is unclear, feel free to issue a PR or discuss further here.

MrMormon:
The only naming information covered there is:

- type names must and the only names to start capitalized (so I removed capitalization from this issue)
- a variable name can have a prime (where, how many?)
- fields can have a leading underscore (no more info than [variables.html](http://tutorial.ponylang.org/expressions/variables.html) gives)

SeanTAllen:
What does "wanting the grammar fully constrained by the tutorial" mean?

MrMormon:
If something is or is not allowed (minus bugs and experimental features), it should be stated or implied by the tutorial. Hence my pedantic questions.

SeanTAllen:
You seem to have a strong grasp on this. Can you PR something that keeps the spirit of the tutorial but covers what you want?

MrMormon:
Actually, sure. Except I don't know the underscore rules.

SeanTAllen:
Awesome. Thanks.

Single underscore means _ private

__ is covered in methods.md:

```
251:underscore followed by a lower case letter. Methods with a leading underscore
253:package. Methods without a leading underscore are public and can be called by
256:__Can I start my method name with 2 (or more) underscores?__ No. If the first
257:character is an underscore then the second one MUST be a lower case letter.
```
__ rule holds true for variables as well.

MrMormon:
Thanks for helping. Double underscores not in the beginning? [This issue](https://github.com/ponylang/ponyc/issues/757)?

SeanTAllen:
No more than 1 underscore at the beginning. 

re: the linked issue, i'd have to look into exactly what the rules are for that. that would be rather low on my priorities right now, if you want to check it out, you can, otherwise, eventually we'll get to that issue.

MrMormon:
Sorry, I meant respectively what about double+ underscores that are not at the beginning of the name, and ending with an underscore?

SeanTAllen:
[...]

PR #104, opened 2016-04-18
Add section: types/classes.md/Naming rules

MrMormon:
First attempt at #99.


SeanTAllen:
thanks @MrMormon

MrMormon:
You're welcome, @SeanTAllen. I hope it's clear that a method means a function, behavior, or constructor. I only implied it by __special__ and __parameter__.

PR #110, opened 2016-05-20
Replace some commas with semicolons

MrMormon:
I think "But be careful, each argument must be specified only once." is fine. Note that I'm American.


MrMormon:
It looks like the rest of the tutorial uses hyphens for these situations. I can close and try that.

jemc:
Yes, hyphens would probably be better for these.

PR #125, opened 2016-06-07
Add documentation on 'embed' fields

georgemarrows:
Inspired by @jemc's explanation on pony+user.


jemc:
I think this could do with some more explanation on the constraint that embedded objects must be constructed when the outer object is constructed, but this explanation you gave so far is already great, so I'm going to merge as-is and the point can be clarified more by someone (whether it's you or someone else) in a future PR.

Thanks for tackling this!

PR #132, opened 2016-06-20
Clarify legal field definitions, minor fixes

0joshuaolson1:
Only I would try to put function definitions before field definitions... This is my best reverse engineered guess about Pony's field grammar rules.

### DIFF
--- a/expressions/variables.md
+++ b/expressions/variables.md
@@ -65,14 +65,15 @@ In Pony, fields are variables that live within objects. They work like fields in
 
 Fields have the same lifetime as the object they're in, rather than being scoped. They are set up by the object constructor and disposed of along with the object.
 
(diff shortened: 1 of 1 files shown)


SeanTAllen:
actors and classes are both types of objects, i don't think the "or actor" is needed.

0joshuaolson1:
My reasoning was that the tutorial seems to equate classes with objects, but I see that [actor literals](http://tutorial.ponylang.org/expressions/object-literals.html) use the object syntax too. I'm canonically wrong, you're right - I'll change it back.

0joshuaolson1:
Today I learned to squash two commits together, yay! Should be better now. All I changed in the embed section is `;` -> `-`

PR #120, opened 2016-05-29
Replace "Wombat" in Rodent interface example

modernserf:
- Wombats are marsupials, not rodents, as they do not satisfy IRL rodents' placental interface. Replaced with Porcupine, an equally large and charismatic critter that is also a true rodent.
- Also, fixed apparent typo in same line: `Array[Rodent` => `Array[Rodent]`

PR #136, opened 2016-06-30
Replace "Rodent" with "Critter" in example

modernserf:
Wombats and capybaras are creatures of similar size, build, and furry-ness, but wombats are marsupials, not rodents. While "mammal" is the closest _biological_ category that encompasses both wombats and capybaras, "critter" is more evocative of their roles in popular culture. (Replaces PR #120)


SeanTAllen:
Damn it @jemc, I REALLY WANTED TO MERGE THE CRITTER FIX. 

DAMN YOU!!!!!

jemc:
:rabbit2:

PR #137, opened 2016-07-02
Change print(count) to print(count.string())

qvks:
Pony does not allow printing a UI32 without .string()


SeanTAllen:
Thanks @qvks !

PR #138, opened 2016-07-10
Add clarifications on how to compile using Docker.

montanonic:
Building from source is something that a fair amount of people (myself included) will always shy away from unless absolutely necessary. Docker is the next available option for Pony, but not everyone has used Docker. This request proposes two minor additions to the installation and build instructions. These additions are targeted towards people who are unfamiliar with Docker, and preempt at least two sources of potential mishap and confusion. Namely: 

* You must use absolute paths
* The absolute path must not end with a trailing slash, `/`

### DIFF
--- a/getting-started/hello-world.md
+++ b/getting-started/hello-world.md
@@ -37,6 +37,8 @@ Writing ./helloworld.o
 Linking ./helloworld
 ```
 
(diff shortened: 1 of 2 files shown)


SeanTAllen:
Thanks!

@montanonic can you open this against the README in ponyc repo? I update from there to here sometimes and if it isn't there, its liable to get lost.

montanonic:
Ah, didn't realize the README had the same content. Yes I can!

SeanTAllen:
Yeah, it's a thing that needs to be addressed eventually. :(

Thanks @montanonic.

PR #139, opened 2016-07-23
Revise the new `gotchas` chapter

mrkishi:
Hello,

This is a simple revision for typos.

Also, I've added a separate commit for each file. Should they be squashed?


SeanTAllen:
Thanks @mrkishi !

PR #140, opened 2016-07-23
Add documentation on equality checks

### LINKED ISSUE #119: Add is/== difference
Many people think that `is` works like `eq` in Java rather than being on identity.

Probably because they see it used like

`if foo is None then`

on primitive types that are singletons.

whereas

`if foo is String then`

will never be true because it instantiates a new String and compares the identity of it to the identity of `foo`. In most instances people want `eq` like if foo == some_string then`

its also possible they are attempting to check the type (this is common as well) and need to use a match statement:

```pony
match foo
| let s: String => ...
else
 ...
end
```

SeanTAllen:
Close #119

### DIFF
--- a/expressions/equality.md
+++ b/expressions/equality.md
@@ -1,0 +1,70 @@
+# Equality in Pony
+
+Pony features two forms of equality: by value and by identity. 
+
+## Identiy equality
+
+Identity equality checks in Pony are done via the `is` keyword. `is` verifies that the two items are the same. Primitives are always equal to 
+
+```pony
+if None is None then
+  // TRUE!
+  // There is only 1 None so the identity is the same
+end
+
+let a = Foo("hi")
+let b = Foo("bye")
+
+if a is b then
+  // NOPE. THIS IS FALSE
+end
+
+let c = a
+if a is c then
+  // YUP! TRUE!
+end
+```
+
+## Value equality
+
+Equality by value check in Pony are done via `==`. They verify that two items have the same value. If the identity of the items being compared is the same then, by definition they have the same value. You can define how equality is defined on your object by implementing `fun eq(that: box->A): Bool`
+
+```pony
+class Foo
+  let _a: String
+
+  new create(a: String) =>
+    _a = a
+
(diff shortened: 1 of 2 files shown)


expressions/equality.md
```diff
@@ -0,0 +1,70 @@
+# Equality in Pony
+
+Pony features two forms of equality: by value and by identity. 
+
+## Identiy equality
+
+Identity equality checks in Pony are done via the `is` keyword. `is` verifies that the two items are the same. Primitives are always equal to
```
jemc: It seems like the last sentence here is unfinished?

expressions/equality.md
```diff
@@ -0,0 +1,70 @@
...
+Identity equality checks in Pony are done via the `is` keyword. `is` verifies that the two items are the same. Primitives are always equal to 
+
+```pony
+if None is None then
+  // TRUE!
+  // There is only 1 None so the identity is the same
+end
+
+let a = Foo("hi")
+let b = Foo("bye")
```
jemc: These two examples would presumably also fail a structural equality check (`==`), since the given string is different for each.  So I don't think it's a good case to demonstrate the difference between structural equality and identity equality (`is`) checking.  It would probably be a good idea to show two objects that are structurally equal, but have different identities.

expressions/equality.md
```diff
@@ -0,0 +1,70 @@
...
+  // NOPE. THIS IS FALSE
+end
+
+let c = a
+if a is c then
+  // YUP! TRUE!
+end
+```
+
+## Value equality
```
jemc: I often here @sylvanc call this "structural equality" - is that a less ambiguous term to use here?

SeanTAllen:
@jemc i made some updates. what do you think?

my current thought is that while I like structural equality as a name, I feel like we might need to be more specific that it is checking by value

Dunno, thoughts?

@aturley ?

jemc:
Thanks! :+1:

SeanTAllen:
looking good @jemc?

PR #146, opened 2016-08-14
Generics documentation

doublec:
This is the start of some documentation for the Generics section of the tutorial. There's a lot more left to do, which I'm still writing but I'd like to push them in smaller pieces so it's not one big review - and I can apply feedback from this to the additional stuff I'm writing.

Please let me know of things to fix up, restructure, add or remove.

### DIFF
--- a/SUMMARY.md
+++ b/SUMMARY.md
@@ -33,7 +33,9 @@
   * [Capability Subtyping](capabilities/capability-subtyping.md)  
   * [Arrow Types](capabilities/arrow-types.md)  
   * [Trust Boundary](capabilities/trust-boundary.md)
-* Chapter 5: Generics
+* [Chapter 5: Generics](generics/index.md)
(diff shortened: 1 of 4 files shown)


SeanTAllen:
This is awesome @doublec.

Should this be squashed or are these logically independent commits?

The only things I see that really should be changed is to format code that same as in other parts of the tutorial:

with the 

<pre>
```pony
whatever
```
</pre>

and each major area should be a H1 not H2 so...

<pre>
# Constraints
</pre>

not 

<pre>
## Constraints
</pre>

Can you make those changes and let me know if I should squash and then I'll merge?

This is an awesome start @doublec. Thank you very much.

doublec:
Thanks @SeanTAllen, I've made the changes and pushed a squashed commit.

SeanTAllen:
AWAY WE GO!

PR #147, opened 2016-08-19
Add a chapter describing Pony's error messages to the appendix

tmmcguire:
This chapter provides examples of code that fails to compile, the error
messages reported by ponyc, a discussion of the error message and the
code that generated it, and suggestions to correct the code.

It would be good to have a general description of Pony's error messages,
but the three examples I started with do not have much consistency. As
a result, I used three related error messages (on mutable/immutable
problems) and went from simple to complex.

The three errors are:

* left side must be something that can be assigned to

* cannot write to a field in a box function

* receiver type is not a subtype of target type


SeanTAllen:
This is awesome @tmmcguire.

Thanks!

Excellent start. Hopefully the community grows this over time and it helps people get past our somewhat difficult error messages (until such time as we have great error messages like Rust or Elm)

PR #149, opened 2016-08-22
Fix markdown link syntax in "Methods" section

leocassarani:
`(Title)[url] -> [Title](url)`


Theodus:
Thanks @leocassarani!

PR #152, opened 2016-08-29
Update installation.md

Triangle345:
Updated windows installation method to reference bintray binary doenlosds


SeanTAllen:
Thanks for this @Triangle345.

Currently we have installation instructions in two places. That is going to be narrowed down to one.
Which looks to be replacing these instructions with a link to the instructions in the GitHub repo.

I'm merging this but it all might disappear soon.

PR #153, opened 2016-09-19
Patches to method documentation

malthe:


expressions/methods.md
```diff
@@ -29,14 +29,16 @@ After the return value there's a `=>` and then finally the function body. The va
...
-class C
-  fun factorial(x: U32): U32 =>
-    if x == 0 then
-	  return 1
-	end
+Pony applies tail call optimization to allow a recursive implementation such as the following factorial function:
 
-	x * factorial(x - 1)
+```pony
+fun factorial(x: I32): I32 ? =>
```
Praetonus: Tail call optimisation doesn't apply here. It is only possible if the recursive call is the very last instruction in the function. Here, the recursive call is followed by a multiplication.

malthe: Is that really true? Isn't it trivial to shift things around so that we can "goto" the beginning of the function?

Praetonus: Sorry, I just checked it, the call is optimised away. But I wouldn't rely on that behaviour for all recursive functions. For example, in a recursive Fibonacci implementation such as the following, the recursive calls aren't eliminated.

```pony
fun fib(n: USize): USize =>
  if n < 2 then
    n
  else
    fib(n - 1) + fib(n - 2)
  end
```

AFAIK, tail call optimisation in LLVM is only guaranteed when the recursive call really is in tail position.

expressions/methods.md
```diff
@@ -8,7 +8,7 @@ __Can I have some code outside of any methods, like I do in Python?__ No. All Po
 
 ## Functions
 
-Pony functions are quite like functions (or methods) in other languages. They can have 0 or more parameters and 0 or 1 return values.
+Pony functions are quite like functions (or methods) in other languages. They can have 0 or more parameters and 0 or 1 return values. A function only returns a value if it declares a return type.
```
malthe: Gotcha – I'll update that.

expressions/methods.md
malthe: If you're writing a function that needs to run recursively an arbitrary number of times, then it has to be tail call optimized – so I think it's important to note that you can and should rely on it in some situations.

It was mentioned on IRC and I thought it should go into the documentation that at least in this case, the optimization applies.

But it would be useful to have the compiler check it, e.g.
```pony
...
else
  inline fib(n - 1)
end
```

Praetonus: I'd say if you want to rely on tail call optimisation, your function should be tail-recursive at the source code level.

A compiler check would be complex to implement. All optimisations of Pony code are made through LLVM and what you're suggesting is non-trivial.

To better convey the idea, I think the code example for this should be a tail-recursive function.

Issue #148, opened 2016-08-21
Link to the type safety proof


kryptan:
Tutorial says:

> It's type safe. Really type safe. There's a mathematical proof and everything.

It would be nice if there was link to the proof right there so people wouldn't need to search it elsewhere.

EpicEric:
Fixed in 7cd8a97.

PR #156, opened 2016-09-20
add proof link

rurban:
Fixes GH #148 

I tried both markdown syntax variants for footnotes [1] and [^1] but it didn't work properly and didn't look good. So I rather went with the simple link.

[1] Sylvan Clebsch and Sophia Drossopoulou. Fully Concurrent Garbage Collection of Actors on Many-Core Machines. [opsla237-clebsch.pdf](https://github.com/ponylang/ponylang.github.io/blob/master/media/papers/opsla237-clebsch.pdf)

This would work though: https://stackoverflow.com/questions/25579868/how-to-add-footnotes-to-github-flavoured-markdown

### DIFF
--- a/docs/getting-started/installation.md
+++ b/docs/getting-started/installation.md
@@ -97,7 +97,17 @@ emerge dev-lang/pony
 
 # Windows
 
-64-Bit installers for Windows 7, 8, 8.1 and 10 will be available soon.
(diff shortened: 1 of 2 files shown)


SeanTAllen:
The installation.md file you modified no longer exists. It was removed a couple weeks back.

You appear to be working against an old version of the tutorial. Any updates for installation needs to be done against ponyc/README.md which is the canonical home.

I'd suggest closing this PR, making sure you are fully up to date with the tutorial and then reapplying anything that might be relevant after that.

I don't think the footnote is needed for the proof. A link to it should suffice.

PR #161, opened 2016-10-23
Remove ffi tuple example

Theodus:
This change reflects the changes about to be made in order to resolve ponylang/ponyc#1240

### DIFF
--- a/c-ffi/calling-c.md
+++ b/c-ffi/calling-c.md
@@ -74,10 +74,17 @@ EGLEvent getEvent() {
     return e;
 }
 ```
-the you can destructure it and get the values using a tuple
+the you can get the value as a struct defined in Pony
 ```pony
-type EGLEvent is (U8, F32, F32)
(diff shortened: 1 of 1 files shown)


jemc:
This isn't quite right.  The `setEvent` example which you changed should actually still work as shown in the current tutorial - we're only talking about disallowing *returning* a pony tuple from an FFI call.

So it's the `getEvent` example (that you didn't change) which will no longer work - it would need to be returning an `EGLEvent*` rather than an `EGLEvent` in order to work.

I recommend that we just remove the `getEvent` example, leave the `setEvent` example as is, and note that FFI calls which return structs by value are not portable enough and consistent enough across compilers to be supported.  Perhaps we should also change the title of "pass C structs" to "pass C structs by value" to emphasize that there is another (non-tuple) mechanism for passing and returning structs by reference.

jemc:
Thank you!

PR #162, opened 2016-10-25
Update break semantics

Praetonus:
A `break` with no value now takes its value from the `else` block instead of being an implicit `None`.


jemc:
Thank you!

PR #166, opened 2016-11-06
Update tutorial to show new lambda syntax.

jemc:
This PR updates the tutorial to show the new lambda syntax. It also makes some corrections around some details about lambdas.

See https://github.com/ponylang/ponyc/pull/1400 for the implementation of the change.

### DIFF
--- a/expressions/object-literals.md
+++ b/expressions/object-literals.md
@@ -37,14 +37,24 @@ class Foo
 
 When we assign to a field in the constructor, we are _capturing_ from the lexical scope the object literal is in. Pretty fun stuff! It lets us have arbitrarily complex __closures__ that can even have multiple entry points (i.e. functions you can call on a closure).
 
-An object literal is always returned as a `ref`, like a default constructor on a class. To get another reference capability (`iso`, `val`, etc.), you can wrap the object literal in a `recover` expression.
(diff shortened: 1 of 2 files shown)


expressions/object-literals.md
```diff
@@ -104,21 +128,22 @@ actor Main
...
+  fun for_each(l: List[String], f: {ref(String)} ref) =>
     try
       f(l.shift())
       for_each(l, f)
     end
 ```
 
-This example declares the type of the apply function that is generated by the lambda expression as being `ref`. The type declaration for it in the `for_each` method also declares it as `ref`. The lambda function captures some variables so the object that is generated is `ref` and the default for the type declaration is `ref` so everything type checks.
+This example declares the type of the apply function that is generated by the 
+lambda expression as being `ref`. The lambda type declaration for the `f` parameter in the `for_each` method also declares it as `ref`. The reference capability of the lambda type must also be `ref` so that the method can be called. The lambda object does not need to declare an explicit reference capability because `val` is the default for a lambda that has captures.
```
Theodus: Shouldn't the last sentence say that `ref` is the default for a lambda that has captures?

jemc: Good catch, it was a copy-paste mistake.  Fixed and squashed.

PR #167, opened 2016-11-10
Add method chaining explanation

Praetonus:


expressions/methods.md
```diff
@@ -196,6 +196,19 @@ class Foo
...
 __Can I call using positional arguments but miss out the first one?__ No. If you use positional arguments they must be the first ones in the call.
 
+## Chaining
+
+Method chaining allows you to chain calls on an object without requiring the method to return its receiver. The syntax to call a method and chain the receiver is `object.>method()`.
+
+```pony
+primitive Printer
+  fun print_two_strings(out: StdStream, s1: String, s2: String) =>
+    out.>print(s1).print(s2)
```
SeanTAllen: so .> chains every call after that?

Praetonus: No, it only chains the next call. `obj.>m()` is roughly equivalent to `(obj.m() ; obj)`. I'll clarify that.

SeanTAllen: Hmm I would have expected to do

`out.print(s1).>print(s2)`

Clarification and more than 1 example would definitely be good.

jemc:
Stylistically, and to avoid confusion, I would suggest recommending:
 `out.>print("1").>print("2").>print("3")`

instead of

`out.>print("1").>print("2").print("3")`

The only difference it makes is in the resulting value of the overall expression.  In the former case, the result is `out`, while in the latter case, the result is the return value of the final `print` call.  Stylistically, we should prefer the continuation of the chain, as it avoids the confusion of mixing `.` and `.>`.

Praetonus:
I've added @jemc's suggestion. The example is indeed way less ambiguous that way, do you think more examples are still required?

SeanTAllen:
Its always good to err on the side of more examples rather than fewer. There's a much better chance over the course of 2,3 or 4 examples that it will click with someone rather than if they only have a single example.

expressions/methods.md
```diff
@@ -196,6 +196,21 @@ class Foo
 
 __Can I call using positional arguments but miss out the first one?__ No. If you use positional arguments they must be the first ones in the call.
 
+## Chaining
+
+Method chaining allows you to chain calls on an object without requiring the method to return its receiver. The syntax to call a method and chain the receiver is `object.>method()`, which is roughly equivalent to `(object.method() ; object)`. Chaining a method discards its original return value.
```
SeanTAllen: I think "original return value" might be hmmm... not sure of the word.

Perhaps "normal return value" ?

expressions/methods.md
```diff
@@ -196,6 +196,21 @@ class Foo
...
+```pony
+primitive Printer
+  fun print_two_strings(out: StdStream, s1: String, s2: String) =>
+    out.>print(s1).>print(s2)
+    // Equivalent to:
+    out.print(s1)
+    out.print(s2)
+```
+
+Note that the last `.>` in a chain can be a `.` if the return value of the last call matters.
```
SeanTAllen: I think it is important to drive home what this means.

If .>print(s2) then the return value is "out".
If .print(s2) then the return value is None

This becomes particularly important for methods that have a non-None value as the normal return value.

Examples of that case as well would be very good to drive home the point of what is happening.

Praetonus: I've added that.

expressions/methods.md
```diff
@@ -196,6 +196,34 @@ class Foo
...
+
+Method chaining allows you to chain calls on an object without requiring the method to return its receiver. The syntax to call a method and chain the receiver is `object.>method()`, which is roughly equivalent to `(object.method() ; object)`. Chaining a method discards its normal return value.
+
+```pony
+primitive Printer
+  fun print_two_strings(out: StdStream, s1: String, s2: String) =>
+    out.>print(s1).>print(s2)
+    // Equivalent to:
+    out.print(s1)
+    out.print(s2)
```
Praetonus: You're right, I forgot to add that when I modified the line with chaining.

PR #172, opened 2017-03-10
Update object-literals.md

sylvanc:
Cover implicit closing over free variables.


expressions/object-literals.md
```diff
@@ -49,6 +49,17 @@ class Foo
     end
 ```
 
+We can also implicitly capture values from the lexical scope by using them in the object literal. Sometimes values that aren't local variables, aren't fields, and aren't parameters of a function are called _free variables_. By using them in a function, we are _closing over_ them - that is, capturing them. The code above could be written withouth the field `s`:
```
jemc: Typo here: `s/withouth/without/`.

SeanTAllen:
This corresponding ponyc change has been merged.

SeanTAllen:
I fixed the typo that @jemc found, squashed and merged.

PR #175, opened 2017-03-13
describe struct usage for C FFI

Charlesetc:
So I used a real-ish example because the part just before this seems to. I'm leaving out the extern declaration and 'main', so the code isn't runnable.. Let me know if there's anything I should add/change! 

The only change in `contributing.md` is reogranized -> reorganized


jemc:
Thanks!

PR #185, opened 2017-04-06
Update location of online standard library

SeanTAllen:
It's moving as part of the move to netlify.


SeanTAllen:
This shouldn't be merged until DNS is updated. Waiting on Sylvan for that.

PR #188, opened 2017-04-08
Update standard library location

ivanov:
This completements the change in ponylang/ponylang.github.io#50 by @SeanTAllen. I assume that the new stdlib location will be consistent, but neither location has the contents of interest at the moment.


ivanov:
Oh, whoops! I failed to check the open PRs, and #185 is identical to this already. Sorry for the noise, closing this.

PR #193, opened 2017-05-03
Update match section to acknowledge exhaustive match.

jemc:
This PR updates the match ~and case function~ sections to acknowledge exhaustive match.

Part of https://github.com/ponylang/ponyc/issues/1772.

Can be merged when the implemented feature is ready to be merged.


jemc:
Wasn't able to implement case methods yet, so altered this PR to only update the docs in the `match` second, and not in the case methods section.

PR #206, opened 2017-07-08
Adding constructor invocation sample

autodidaddict:
When going through the tutorial myself, I had to switch out and google for the syntax to invoke the named constructor. Thought it would be handy to put this in here in the tutorial.


SeanTAllen:
Thanks @autodidaddict. Welcome!

PR #208, opened 2017-07-11
Types: behaviours now return None

### LINKED ISSUE #196: Description of what a behavior returns is outdated
Now that behaviors return `None`, this information in `Types/Actors` is no longer accurate:

```
So what does a behaviour return? All behaviours always return the receiver. They can't return something they calculate (since they haven't run yet), so returning the receiver is a convenience to allow chaining calls on the receiver.
```

jmiven:
Since RFC 28 and v0.10.0, all behaviours return None instead of the receiver.

Fixes #196

### DIFF
--- a/types/actors.md
+++ b/types/actors.md
@@ -12,7 +12,7 @@ A behaviour looks like a function, but instead of being introduced with the keyw
 
 Like a function, a behaviour can have parameters. Unlike a function, it doesn't have a receiver capability (a behaviour can be called on a receiver of any capability) and you can't specify a return type.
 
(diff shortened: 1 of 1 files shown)


jemc:
I'd also tend to not mention the old semantics, but I'd be fine with leaving them in if someone feels strongly about it being helpful.

jmiven:
I included the details on how it worked before because I was myself surprised by the change (I stopped following Pony for several months).

I realise now that they indeed aren't helpful for most (new) readers, so I'll remove them.

PR #215, opened 2017-08-04
Clarify confusing paragraph about structural equality

non-Jedi:
I was experiencing confusion over which side of of `a == b` called `eq`, so I made the section on structural typing easier for me to understand.


SeanTAllen:
Thanks @non-Jedi!

PR #221, opened 2017-08-18
Document existing annotations

Praetonus:


jemc:
@Praetonus - Do you think it would be helpful to add an example or two to demonstrate the correct Pony syntax for where the annotation should appear? I know that you described them appearing "after any scoping keyword or symbol", but I wonder if an example or two would help some folks to visualize that.

Praetonus:
@jemc I'll add examples for the recognised annotations.

SeanTAllen:
@jemc does this look good to you?

Issue #222, opened 2017-08-23
Add section about literals in general and string literals in particular


mfelsche:
During discussion of https://github.com/ponylang/ponyc/issues/2178 during the sync i volunteered to add some section about literals to the tutorial,
as the indent removal of triple quoted strings needs some explanation as it might be irritating to users.

mfelsche:
This was fixed by #223

PR #223, opened 2017-08-27
Add section about literals

mfelsche:
it covers boolean, numeric, character, string and array literals.

Please consider this as a suggestions and feel free to cut, scramble and criticize my writing.
I take no pride from my writing, so roast me. :)

### Concerns
* it might go a little bit too much into detail on array literals and type inference
* it says string literals can contain any unicode characters (which works for me (UTF-8 source file, pony playground) but was reported from several sources that pony does not)

### DIFF
--- a/expressions/literals.md
+++ b/expressions/literals.md
@@ -1,0 +1,299 @@
+# Literals
+
+***What do we want?***
+
+Values!
+
+**Where do we want them?***
+
+In our pony programs!
+
+**Say no more**
+
+Every programming language has literals to encode values of certain types,
+and so does Pony.
+
+In pony you can express booleans, numeric types, characters, strings
+and arrays as literals.
+
+## Bool Literals
+
+There is ``true``, there is ``false``. That's it.
+
+## Numeric Literals
+
+Numeric literals can be used to encode any signed or unsigned integer
+or floating point number.
+
+In most cases pony is able to infer the concrete type of the literal from
+the context where it is used (Assignment to a field or local variable,
+as argument to a method/behavior call).
+
+It is possible to help the compiler determine the concrete type of the literal
+using a constructor of one of the numeric types:
+
+ * U8, U16, U32, U64, U128, USize, ULong
+ * I8, I16, I32, I64, I128, ISize, ILong
+ * F32, F64
+
+```pony
+let my_explicit_unsigned: U32 = 42_000
+let my_constructor_unsigned = U8(1)
+let my_constructor_float = F64(1.234)
+```
+
+Integer literals can be given as decimal, hexadecimal
+or binary:
+
+```pony
+let my_decimal_int: I32 = 1024
+let my_hexadecimal_int: I32 = 0x400
+let my_binary_int: I32 = 0b10000000000
+```
+
+Floating Point literals are expressed as standard floating point 
+or scientific notation:
+
+```pony
+let my_double_precision_float: F64 = 0.009999999776482582092285156250
+let my_scientific_float: F32 = 42.12e-4
+```
+
+## Character Literals
+
+Character literals are enclosed with single quotes (``'``).
+
+They are used to encode single byte characters 
+but can be coerced to any numeric type:
+
+```pony
+let big_a: U8 = 'A'
+let hex_escaped_char: U8 = '\x41'
+let newline: U32 = '\n'
+```
+
+The following escape sequences are supported:
+
+* ``\x4F`` hex escape sequence for unicode letters with 2 hex digits (up to 0xFF)
+* ``\a``, ``\b``, ``\e``, ``\f``, ``\n``, ``\r``, ``\t``, ``\v``, ``\\``, ``\0``, ``\"``
+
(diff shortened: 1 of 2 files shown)


SeanTAllen:
I didn't check that everything works as said but I think this is pretty good.

Could you run it through a free grammar checker like grammarly.com and address any issues.

Otherwise, I'm good with this.

mfelsche:
Thanks for the review. I updated the section and fixed some typos.
All source examples are playground-tested.
Good to go, from my side.

mfelsche:
This fixes #222 but it should not be published before https://github.com/ponylang/ponyc/issues/2178 as it describes behavior for triple quoted strings that is not yet implemented.

expressions/literals.md
```diff
@@ -0,0 +1,319 @@
...
+
+To safely create an array literal of type ``Array[T]`` it is necessary
+to use elements of type ``T^``, an ephemeral type of ``T``;
+A ``T`` that is not assigned to any variable yet.
+"Consuming" a variable returns an ephemeral type,
+all constructors return ephemeral types, and other literal expressions work
+the same:
+
+```pony
+let ref = U64(123)
```
jemc: You can't use `ref` as a variable name because it is a reserved word.
Please make sure all your examples compile, because if they're broken they'll be very confusing to newcomers, who won't know how to fix them.

expressions/literals.md
```diff
@@ -0,0 +1,319 @@
...
+let safe_array: Array[U64 val] =
+  [
+    consume ref     // consuming a variable
+    U64.create(1)   // constructor
+    U64.min_value() // constructor
+  ]
```
jemc: Your array style is inconsistent throughout this page. Please check the style guide in the ponyc repo for the style that is used in the standard library and should be adopted across documentation as well.

expressions/literals.md
```diff
@@ -0,0 +1,319 @@
...
+A ``T`` that is not assigned to any variable yet.
+"Consuming" a variable returns an ephemeral type,
+all constructors return ephemeral types, and other literal expressions work
+the same:
+
+```pony
+let ref = U64(123)
+let safe_array: Array[U64 val] =
+  [
+    consume ref     // consuming a variable
```
jemc: A `U64 val` is not a good example of consuming something, because `val` has no uniqueness constraints.

expressions/literals.md
```diff
@@ -0,0 +1,319 @@
...
+    U64.create(1)   // constructor
+    U64.min_value() // constructor
+  ]
+```
+
+Most of the time constructing array literals is not a problem,
+especially when using only literal elements (Strings, Numbers).
+These have the reference capability ``val`` which is safe to alias multiple
+times and whose aliases are safe to also be ``val``.
```
jemc: Only one backtick is needed on each side surround a keyword name in markdown, not two.

expressions/literals.md
```diff
@@ -0,0 +1,319 @@
...
+  "A"
+  "B"
+  "C"
+]
+```
+
+Because it is equivalent to this:
+
+```pony
+let explicit_safe_array: Array[String val!] = [
```
jemc: Again, `val!` is not a good example, because `val` has no uniqueness constraints.

expressions/literals.md
```diff
@@ -0,0 +1,319 @@
...
+
+Consider the following example:
+
+```pony
+// does not compile
+let s: String iso = recover iso String(0) end
+let array: Array[String iso] = [s]
+```
+
+*It does not work. Why?*
```
jemc: I do think that this is going into too much detail about this aspect of arrays.

This is not really a property of array literals so much as it being a property of capabilities. If your capability has uniqueness constraints (`iso` or `trn`), then you can't hold it as such in two places at once, including the particular case where one of those places is in an array, but note that the array isn't really a special case for this at all.

I think we can describe this succinctly in a sentence or two without going through a full worked example about it.

expressions/literals.md
```diff
@@ -0,0 +1,319 @@
...
+```pony
+// does not compile
+let my_stringable_array: Array[Stringable] =
+  [
+    U64(0xA)
+    "0xA"
+  ]
+```
+
+### Arrays and References
```
jemc: I do think that this is going into too much detail about this aspect of arrays.

This is not really a property of array literals so much as it being a property of capabilities. If your capability has uniqueness constraints (`iso` or `trn`), then you can't hold it as such in two places at once, including the particular case where one of those places is in an array, but note that the array isn't really a special case for this at all.

I think we can describe everything below this header succinctly in a sentence or two without going through full worked examples about it.

mfelsche: I shortened this section to only mention that element types are inferred as their aliases.

mfelsche:
this is good to go from my side, now that https://github.com/ponylang/ponyc/issues/2178 is merged

mfelsche:
I had to correct the section as array type inference was released with 0.19.1

SeanTAllen:
@jemc this look good to you?

PR #224, opened 2017-08-27
clarify that string literals don't need to be ASCII-only

mfelsche:
Correctly and concisely stating what source code elements need to be ASCII may need some tweaking.

### DIFF
--- a/types/classes.md
+++ b/types/classes.md
@@ -159,7 +159,9 @@ We'll talk about all that stuff in detail later.
 
 ## Naming rules
 
-By now it shouldn't be very surprising to learn that Pony is written in __ASCII__. [ASCII](https://en.wikipedia.org/wiki/ASCII) is a standard text encoding that uses English characters and symbols, and almost every programming language in existence defines source code as a subset of it.
+By now it shouldn't be very surprising to learn that in Pony identifiers such as class-, method- and variable-names and other source code elements may only contain [__ASCII__](https://en.wikipedia.org/wiki/ASCII).
+
+String literals are the exception to the rule. They can contain any byte sequence preferred, although Pony ``String``s cooperate best with UTF-8 (e.g. ``String.codepoints()`` for counting the number of [Unicode](https://en.wikipedia.org/wiki/Unicode) codepoints requires UTF-8 to work correctly).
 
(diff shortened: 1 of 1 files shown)


SeanTAllen:
I think this is a little unclear. I don't think I would understand from this. That said, I'm not sure the best way to express. @jemc any thoughts?

mfelsche:
you are right. Let me give it another try.

mfelsche:
My thoughts behind this one: I limited the part about ASCII to names/identifiers. As this is the only "user-generated" content of a pony program for which the required character-set might be relevant. All Pony keywords and other constructs are a fixed set, which just happens to be ASCII, so mentioning them is not necessary.

Some people fell for the trap and believed that pony code needs to be completely ASCII-only, which is not the case. That is why i added the part about String literals. As they support any kind of characters.

jemc:
Honestly, I'm not really sure that other encodings for string literals are *properly* supported in any real sense. As far as I know, it was not really written with the intention of supporting them, and the extent to which it does support them is mostly "by accident", and there are likely some edge cases (maybe related to escape sequences or other non-literal treatment) that don't behave as expected.

String literals are just sequences of bytes, read directly from the bytes written in the file, with a few escape sequences getting replaced by other bytes. They have no intrinsic encoding. If you have bytes that your editor happens to display a particular set of bytes as a smiley-face, and you have a terminal that will display that same set of bytes as the same smiley-face, then you can print that string literal and probably expect to see it in your output. However, Pony is just seeing a sequence of bytes - it's not "supporting that encoding" - it's just reading the bytes in the file and printing those same bytes to STDOUT - give or take some escape sequences, which complicate the question a bit more.

Also, you have to be careful about using the word "character" when talking about string encodings, as some people expect it to mean "bytes" and others expect it to refer to a multi-byte sequence that is visually displayed as a single character (more properly called a "rune").

mfelsche:
I tried to get the wording straight to ensure it is understood correctly that String literals are not limited to be ASCII but handle given source bytes transparently.

SeanTAllen:
@jemc does this look good to squash and merge?

jemc:
Made a few adjustments, and will merge. Sorry for holding this up.

mfelsche:
Thank you!

PR #229, opened 2017-09-28
Adding example to concurency and sequential actor story

xcombelle:
I just discovered about Pony. And reading the tutorial, what is concurrent and sequential was not clear. It appeared to me that adding example, small rewording and implementation detail make it clearer.

The example is in two fold: first that due to concurrency, interleaving might happen and second that due to sequential character, the order of execution will be the same that the order of calls

### DIFF
--- a/types/actors.md
+++ b/types/actors.md
@@ -30,13 +30,59 @@ Here we have an `Aardvark` that can eat asynchronously. Clever Aardvark.
 
 ## Concurrent
 
-Since behaviours are asynchronous, it's ok to run the body of a bunch of behaviours at the same time. This is exactly what Pony does. The Pony runtime has its own scheduler, which by default has a number of threads equal to the number of CPU cores on your machine. Each scheduler thread can be executing an actor behaviour at any given time, so Pony programs are naturally concurrent.
+Since behaviours are asynchronous, The runtime might execute a bunch of behaviours at the same time. This is exactly what Pony does. The Pony runtime has its own scheduler, which by default has a number of threads equal to the number of CPU cores on your machine. Each scheduler thread can be executing an actor behaviour at any given time, so Pony programs are naturally concurrent.
+
+for example, the following code:
+
+```
+actor A
+  var name:String
+  var env:Env
+  new create(env':Env,name':String) =>
+    name = name'
+    env = env'
+
+  be first() =>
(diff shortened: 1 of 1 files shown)


types/actors.md
```diff
@@ -30,13 +30,59 @@ Here we have an `Aardvark` that can eat asynchronously. Clever Aardvark.
...
+
+```
+might have the following resul:
+
+```
+b first
+a first
+b second
+a second
+```
```
aturley: Your example result isn't actually possible in Pony, as you point out below. Therefore I worry that this section is misleading and could confuse someone. I don't think there's any need to introduce this level of potential confusion here.

types/actors.md
```diff
@@ -30,13 +30,59 @@ Here we have an `Aardvark` that can eat asynchronously. Clever Aardvark.
...
 
 ## Sequential
 
 Actors themselves, however, are sequential. That is, each actor will only execute one behaviour at a time. This means all the code in an actor can be written without caring about concurrency: no need for locks or semaphores or anything like that.
 
-When you're writing Pony code, it's nice to think of actors not as a unit of parallelism, but as a unit of sequentiality. That is, an actor should do only what _has_ to be done sequentially. Anything else can be broken out into another actor, making it automatically parallel.
+When you're writing Pony code, it's nice to think of actors not as a unit of parallelism, but as a unit of sequentiality. That is, the methods (function, behavior, constructor and finaliser) are exectuted in the order they are called. Anything else can be broken out into another actor, making it automatically parallel.
```
aturley: > That is, the methods (function, behavior, constructor and finaliser) are exectuted in the order they are called

This isn't really true. If a behavior in the same actor is called, and then a function is called, the function will be executed first, then later, once the currently executing behavior has exited, the behavior call will execute.

xcombelle: @aturley, so only behavior are guarantee to be in order as called ? and functions are called immediatly ?

SeanTAllen: correct. functions are synchronous and behaviors are asynchronous.

mfelsche:
@xcombelle what is the status of this PR? If you are blocked by something, can we help to get your work rolling again?

SeanTAllen:
I'm closing this as its been open for a very long time without any updates.

PR #236, opened 2017-10-14
Add a small example of behaviours in methods.md

codec-abc:


expressions/methods.md
```diff
@@ -228,6 +228,47 @@ primitive Foo
     f.>add_option(o1).>add_option(o2).make_object() // Works. The expression returns an Object
 ```
 
+## Behaviours
+Behaviours are asynchronous method calls. As such, they cannot return anything, and the caller will not wait for the call to finish. Here is a quick example to illustrate this certainly :
+
+```pony
+/** 
+This program will certainly print the following :
```
mfelsche: I am not 100% sure that this is always the case. afaik it could be that `OtherActor.ping` is called before the last print in `Main.create` is evaluated. But i can be wrong about this.

expressions/methods.md
```diff
@@ -228,6 +228,47 @@ primitive Foo
     f.>add_option(o1).>add_option(o2).make_object() // Works. The expression returns an Object
 ```
 
+## Behaviours
+Behaviours are asynchronous method calls. As such, they cannot return anything, and the caller will not wait for the call to finish. Here is a quick example to illustrate this certainly :
```
mfelsche: > Behaviours are asynchronous method calls.

To be precise (i think we need to be here to avoid confusion), behaviours are not method calls, but methods that are asynchronously evaluated in the context of their containing actor, that is possibly on another thread.

Maybe it is worth mentioning that they also cannot error.

SeanTAllen:
I dont think this is appropriate as is here. I think it should be pointing back to discussion of behaviors in the Actor intro.

codec-abc:
Make sense. I think the exemple could even be reworked to show how data can be passed between actors. Maybe having an actor spawning a new one to compute some simple math.

mfelsche:
@codec-abc what is the status of this? should this be closed or are you planning on working on this?

codec-abc:
Since everybody does not agree on putting it here, I think it is better to close it the issue.

PR #237, opened 2017-10-14
Add a link to the playground in the getting started section.

codec-abc:


getting-started/what-you-need.md
```diff
@@ -1,6 +1,6 @@
 # What You Need
 
-To get started, you'll need a text editor and the [ponyc](https://github.com/ponylang/ponyc) compiler. That's it.
+To get started, you'll need a text editor and the [ponyc](https://github.com/ponylang/ponyc) compiler. Or if you are on a not supported platform or don't want to install the compiler you can use the [Pony's Playground](https://playground.ponylang.org/)
```
mfelsche: The way you motivate mentioning the playground, one could also mention the `ponylang/ponyc` docker container. What do you think?

codec-abc: Why not. The advantage of the playground is that it can be used by anyone. How would advertise the container? Adding a link to the pony repository?

SeanTAllen:
My worry with this is that you are going to eventually have all this info across more than one location. I'd rather not add the docker container here, better to direct people to where the INSTALL options are. If they aren't good enough there. They should be fixed there.

Playground? Yes.
Docker container? I think that's a step to far.

getting-started/what-you-need.md
plietar: The first link points to how to install, which includes how to use docker. So I don't think it is necessary here

PR #239, opened 2017-10-18
use gitbook plugin to add button to run code sample

codec-abc:
This add a GitBook plugin to add a button to send code samples to the pony playground. There is not much to see in this request because it just add a GitBook plugin that can be seen [here](https://github.com/codec-abc/gitbook-plugin-copy-code-button/blob/master/book/toggle.js). I used [this plugin](https://github.com/WebEngage/gitbook-plugin-copy-code-button) as a base for my own. Basically it works the following way:

It find each code sample in the page and a button near it. When the button is clicked, I retrieve the code in the sample (in a ugly way I think, but I am no Js expert) created a gist of it -using Github API- and finally redirect the page to the playground with the newly created gist.

The drawbacks are the following:
* It change the current page URL (While reader will probably want to open it in a new tab/page. It can be changed easily but I think most browser will block this behavior by default)
* It add a button for every sample no matter for the language.
* Because most samples are not strictly valid (often lacking the ```actor Main``` declaration they will not compile as they are, reducing the utility of this tool.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,5 @@ site/
 .DS_Store
 *.sublime-project
 *.sublime-workspace
+_book/
+node_modules/
--- a/book.json
+++ b/book.json
@@ -3,7 +3,7 @@
         "footnote-string-to-number",
         "highlight",
         "lunr@1.2.0",
-        "run-code-button-pony"
+        "run-code-button-ponylang"
     ],
     "pluginsConfig": {
         "lunr": {
--- a/book.json
+++ b/book.json
@@ -2,6 +2,7 @@
     "plugins": [
         "footnote-string-to-number",
         "highlight",
+        "lunr@1.2.0",
         "run-code-button-pony"
     ],
     "pluginsConfig": {
--- a/book.json
+++ b/book.json
@@ -2,7 +2,7 @@
     "plugins": [
         "footnote-string-to-number",
         "highlight",
-        "lunr@1.2.0"
+        "run-code-button-pony"
     ],
     "pluginsConfig": {
         "lunr": {


jemc:
We talked about it a bit on the pony sync call.

We really like the idea, but are concerned about the "most samples are not strictly valid" caveat you mentioned.

I took a quick look at the source code for the plugin, and I'm thinking that a reasonably simple change would be to check for the string "actor Main" in the HTML, and only do the expand button if that string is present. It's unlikely that incomplete examples would have this string, but in that unlikely case that they do, we can fix them on a case by case basis.

Once we had that change in place, we'd be very happy to accept this change.

codec-abc:
I update the code of the GitBook's plugin to reflect the changes wanted. I was able to detect easily the code sample language so it will show only on code sample that specify that they are written in Pony. Plus, it now checks that the sample contains at least the strings "actor Main" and "new create(env: Env) =>". With that changes it only shows on 2 code samples that are the hello world ones that are the 2 last entries in chapter 1. One thing though, do you want to fork my plugin repository so the pony organization can have ownership of the NPM package for the gitbook plugin?

jemc:
> One thing though, do you want to fork my plugin repository so the pony organization can have ownership of the NPM package for the gitbook plugin?

I think that sounds like a good idea, if you're willing.

jemc:
I've forked the repo to ponylang - can you give further instructions on what else is required to make this work?

codec-abc:
It is quite simple. Obviously you will need a NPM account and NodeJs installed on your computer. I don't know if you plan to have a one that you share between core members or each core member has its own. I will assume that you only want to use a single account for the sake of simplicity. But NPM has an organization feature similar to github, so I think you can start with one account and then create and manage an organization if needed.

To publish the package on NPM you need to do the following:
* Open the package.json and edit the fields:
  * name (otherwise it will clash with the existing one)
  * homepage
  * repository.url
  * author
  * _npmUser.name
  * bugs.url
* Commit and push the change
* type `npm adduser` and follow instructions
* type `npm publish` and follow instructions

At this point your package should be uploaded on NPM. You just have to update the book.json at this [line](https://github.com/ponylang/pony-tutorial/pull/239/files#diff-609ea0a435d067757e821e3114b666a1R6) to use the name of your fork.

jemc:
Awesome, let's give it a try!

jemc:
Seems to be working.

Thanks for doing this!

codec-abc:
No problem. I am gonna do a pass on all the examples, because a few of them now have the button but don't compile (and it is generally not intended).

PR #240, opened 2017-10-21
Add samples for the standard library.

codec-abc:


packages/standard-library.md
```diff
@@ -5,3 +5,173 @@ The Pony standard library is a collection of packages that can each be used as n
...
+
+Obviously, the regex package deals with regular expressions.
+
+```pony
+use "regex"
+
+actor Main
+  new create(env: Env) =>
+    try
+      let r = Regex("pony is .*")?
```
mfelsche: what about showing some grouping features here?

e.g.:

```pony
let r = Regex("(pony is) .*")?
let result = r.replace("pony is a kind of horse", "$1 awesome")?
env.out.print("result.clone())
```

packages/standard-library.md
```diff
@@ -5,3 +5,173 @@ The Pony standard library is a collection of packages that can each be used as n
...
+      let result = r.replace("pony is a kind of horse", "awesome")?
+      env.out.print("pony is" + result.clone())
+    end
+```
+
+### The file package
+
+The file package is pretty much what you expect. It deals with files and directories.
```
mfelsche: maybe also mention paths to files.

packages/standard-library.md
```diff
@@ -5,3 +5,173 @@ The Pony standard library is a collection of packages that can each be used as n
...
+    _env.out.print(_counter.string())
+    _counter = _counter + 1
+    if (_counter > 10) then
+        false
+    else
+        true
+    end
+```
+
+### Others packages
```
mfelsche: typo: `Other`

packages/standard-library.md
```diff
@@ -5,3 +5,173 @@ The Pony standard library is a collection of packages that can each be used as n
...
+
+This is enough for an introduction but as you will explore Pony you may find other packages really helpful.
+The different net packages are used for anything that is network related.
```
mfelsche: maybe add a link like this: [`net`](https://stdlib.ponylang.org/net--index/)

packages/standard-library.md
```diff
@@ -5,3 +5,173 @@ The Pony standard library is a collection of packages that can each be used as n
...
+The promise package will help you collect result of work done by others actor. The options one will help you parse and manage arguments from the command line.
+The random package allow you to play Russian roulette. The crypto package contains the common hashing function.
```
mfelsche: please also add links to stdlib docs for those packages.

mfelsche: typo: `The crypto package contains the common hashing functions.`

packages/standard-library.md
```diff
@@ -5,3 +5,173 @@ The Pony standard library is a collection of packages that can each be used as n
...
+    var inhabitants_count_per_country = HashMap[String, F32, StringHasher]
+    
+    inhabitants_count_per_country.update("France" , F32(70_000_000))
+    inhabitants_count_per_country.update("USA" , F32(323_000_000))
+    
+    for (key, value) in inhabitants_count_per_country.pairs() do
+      env.out.print(key + " has " + value.string() + " inhabitants")
+    end
+```
+
```
mfelsche: Maybe also at least mention the `collections.persistent` package.

codec-abc: Just to make sure, the difference between the 2 packages are that when you do an operation on a collection from collections.persistent it is making a copy of the old and return a new one with the operation applied?

packages/standard-library.md
codec-abc: neat

SeanTAllen:
Thanks for doing this, however, it's not something I'm comfortable with. At a recent sync we discussed having an overview of the standard library in the tutorial but no "how to use package X". That "how to use" should be in the package documentation itself.

I'm comfortable with a "these are the packages in standard library." I think any "this is how to use them" should point to the relevant documentation for the standard library. Not exist here.

packages/standard-library.md
```diff
@@ -5,3 +5,175 @@ The Pony standard library is a collection of packages that can each be used as n
...
+        false
+    else
+        true
+    end
+```
+
+### Other packages
+
+This is enough for an introduction but as you will explore Pony you may find other packages really helpful.
+The different net packages - [net](https://stdlib.ponylang.org/net--index) [net/http](https://stdlib.ponylang.org/net-http--index) and [net/ssl](https://stdlib.ponylang.org/net-ssl--index) -  are used for anything that is network related.
```
mfelsche: `net/http` has been removed from the stdlib

packages/standard-library.md
```diff
@@ -5,3 +5,175 @@ The Pony standard library is a collection of packages that can each be used as n
...
+The [promise](https://stdlib.ponylang.org/promises--index) package will help you collect result of work done by others actor. The [options](https://stdlib.ponylang.org/options--index) one will help you parse and manage arguments from the command line.
```
mfelsche: the `options` package is actually already deprecated in favor of the `cli` package.

mfelsche:
@SeanTAllen you comfortable with this now after the latest change? I do think this has a place in this tutorial. Thunderdome?

SeanTAllen:
I think this belongs with the documentation for the standard library. Not the tutorial.

PR #243, opened 2017-10-27
Make the type conversion example compile

codec-abc:


SeanTAllen:
@codec-abc I made a couple small changes for concision. Thanks!

PR #245, opened 2017-11-01
Rework "exceptions"

SeanTAllen:
Pony's error isn't the same as exceptions in other languages. This is
the first pass at addressing.


jemc:
Should we go ahead and just rename the section to "errors" or "runtime errors", getting rid of all instances of the word "exception" as referring to Pony?

SeanTAllen:
I was thinking "Exceptions" stays purely so folks can find it easier when looking. But, I'm not sure its the correct thing to do.

I could go either way.

PR #246, opened 2017-11-05
Update object-literals.md

surma:
The section this PR affects didn’t quite make sense to me when I read it. Looking at the [last change](https://github.com/ponylang/pony-tutorial/commit/e9d831e30651a31bc3f80313097f780935cb90ec#diff-5d9038176c31de2ad76e34b1b117f132L82) to that section, I feel like the code example was added by accident.

PTAL :)

### DIFF
--- a/expressions/object-literals.md
+++ b/expressions/object-literals.md
@@ -90,18 +90,7 @@ object iso
 end
 ```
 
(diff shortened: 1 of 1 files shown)


mfelsche:
good catch, thank you!

PR #247, opened 2017-11-06
Update recovering-capabilities.md

surma:
I think the explanation of default reference capabilities should be moved up a bit as I find it confusing to read “That's a simple one” and I wasn’t able to understand why the result is `Array[String] iso`.


mfelsche:
great, thanks!

PR #248, opened 2017-11-11
Fix multiply function definition formatting

natdempk:
This is an incredibly minor fix for a small formatting issue I notice while going through the tutorial.

`fun multiply(a : U64, b: U64) : U64` should be `fun multiply(a: U64, b: U64) : U64` instead so the formatting is consistent.

### DIFF
--- a/types/primitives.md
+++ b/types/primitives.md
@@ -28,7 +28,7 @@ primitive BasicMath
   fun add(a: U64, b: U64) : U64 =>
     a + b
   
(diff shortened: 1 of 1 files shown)


SeanTAllen:
I'm going to merge this because its an improvement. However all those ") : U64" are wrong and not what is in the style guide.

`fun multiply(a: U64, b: U64) : U64` 

should be

`fun multiply(a: U64, b: U64): U64"

If anyone wants to address those across the tutorial, that would be great.

Thanks for the PR @natdempk!

PR #249, opened 2017-11-12
Fix formatting for function return types

natdempk:
This fixes the formatting issue mentioned here: https://github.com/ponylang/pony-tutorial/pull/248#issuecomment-343697691

Running `ag "\)\s+:\s*"` across the tutorial only showed these two results with incorrect formatting, so seems like this incorrect whitespace isn't an issue anywhere else.


SeanTAllen:
Thanks @natdempk. You rock!

PR #250, opened 2017-11-13
Cleanup references to quotes

winksaville:
The reference in String Literals to a single quote is wrong and changed
triple quote to triple-quoted to be consitent with other uses in the
documentation.


mfelsche:
Thanks a million, it now sounds much better.

PR #256, opened 2017-11-30
Add side effect ordering in function calls gotcha

### LINKED ISSUE #158: Add https://github.com/ponylang/ponyc/issues/863 to gotchas
https://github.com/ponylang/ponyc/issues/863

SeanTAllen:
Closes #158

### DIFF
--- a/gotchas/side-effect-ordering-in-function-call-expressions.md
+++ b/gotchas/side-effect-ordering-in-function-call-expressions.md
@@ -17,4 +17,4 @@ actor Main
 
 What do you think it will print? Probably `0` right? Or maybe you realized this code is in the gotchas section so it must be `42`. If you went with `42`, you'd be right. Why?
 
-Expressions for arguments in function calls  are evaluated before the expression for the function object itself. The use of assignment expressions like `x = 42` is quite rare so we don't think many folks will be bitten by this, however, its definitely something you want to be aware of.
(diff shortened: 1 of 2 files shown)


gotchas/side-effect-ordering-in-function-call-expressions.md
```diff
@@ -0,0 +1,20 @@
...
+    var x: U64 = 0
+    try foo()?.fn(x = 42) end
+    env.out.print(x.string())
+
+  fun foo(): Foo ? => error
+```
+
+What do you think it will print? Probably `0` right? Or maybe you realized this code is in the gotchas section so it must be `42`. If you went with `42`, you'd be right. Why?
+
+Expressions for arguments in function calls  are evaluated before the expression for the function object itself. The use of assignment expressions like `x = 42` is quite rare so we don't think many folks will be bitten by this, however, its definitely something you want to be aware of.
```
jemc: I think the phrasing of "the expression for the function object itself" is a bit confusing - I had to read it a few times before understanding what it meant. I'd be happier with calling the the "the expression for the function receiver" - or we could be even more explicit and say something like "the expression that the function is being called on".

jemc: Also there's an extra space after "function calls" here.

SeanTAllen:
@jemc do you want to make an edit to make it more clear and then merge?

PR #257, opened 2017-11-30
Remove early reference to "error" in tail call example

### LINKED ISSUE #183: The "error" in factorial function in Methods chapter is not described
It would be nice to have a reference to error handling docs after the factorial function example. Without it the "if x < 0 then error end" line looks like error :-).

SeanTAllen:
Closes #183

### DIFF
--- a/expressions/methods.md
+++ b/expressions/methods.md
@@ -32,13 +32,15 @@ If you want to exit a function early then use the `return` command. If the funct
 Pony applies tail call optimization when applicable to allow a recursive implementation such as the following factorial function:
 
 ```pony
-fun factorial(x: I32): I32 ? =>
-  if x < 0 then error end
-  if x == 0 then
-    1
-  else
-    x * factorial(x - 1)?
-  end
+primitive Factorial
+  fun factorial(x: I32): I32 =>
+    if x < 0 then 
+      0
+    elseif x == 0 then
+      1
+    else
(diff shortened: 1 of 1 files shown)


expressions/methods.md
```diff
@@ -32,13 +32,15 @@ If you want to exit a function early then use the `return` command. If the funct
...
 ```pony
-fun factorial(x: I32): I32 ? =>
-  if x < 0 then error end
-  if x == 0 then
-    1
-  else
-    x * factorial(x - 1)?
-  end
+primitive Factorial
+  fun factorial(x: I32): I32 ? =>
```
aturley: Since the function can no longer raise an error it no longer needs to be partial.

SeanTAllen:
odd that it compiles.

SeanTAllen:
@aturley updated.

aturley:
@SeanTAllen I filed #2396 in response to the fact that your partial version compiled even though the function wasn't really partial.

SeanTAllen:
@aturley i had already opened an issue for that and closed #2396 as a duplicate.

SeanTAllen:
@jemc updated to U32. merging.

SeanTAllen:
Apparently I forgot to actually push the U32 change. need to wake up more.

SeanTAllen:
Fixing on master.

PR #260, opened 2017-12-19
Mention --bin-name option to set the executable binary name.

winksaville:


winksaville:
If [the change](https://github.com/ponylang/ponyc/pull/2430) which add --bin-name is merged, then mentioning that will be helpful.

PR #261, opened 2017-12-23
Character literal escape sequence correction

ii8:


mfelsche:
Thanks, this was indeed a small nasty error, i introduced here. :)

PR #269, opened 2018-02-06
match.md: Fix examples for matches that don't compile.

jghoman:
Two examples in match.md do not compile with the either 0.21.3 or the latest versions of ponyc.

The first example, an else, fails because the other matches are exhaustive and therefore else is unreachable.  This else branch is deleted in this patch.

The second example re-uses the variable named y.  This variable is renamed to u in this patch.


mfelsche:
Thank you very much! Good catch!

PR #270, opened 2018-02-06
multiple traits

gwelr:
describing how to combine multiple traits together, with an accent on the syntax.


mfelsche:
Thank you very much, i actually had some moments where this would have been very useful already.

PR #272, opened 2018-02-21
Remove "divide by zero" gotcha

kt3k:
I checked with the current ponyc and I confirmed that the behavior mentioned in this page doesn't seem existing now. (If the number is U64, `1/0` doesn't seem compiling. If the number is F64, `1/0` seems returning `inf`.)

This behavior seems to have changed at https://github.com/ponylang/ponyc/pull/1395. I'm afraid this gotcha page is now misleading.


SeanTAllen:
Hi @kt3k. This is still valid. Floats will return inf as that is the standard. U64 will not, however, 1/0 as constants will result in a compiler error but that's a really rare circumstance. If you were to supply 0 for division via almost any other means you will get 0.

See https://playground.ponylang.org/?gist=8b1d2b7f193af4b1d7b3a38924e9f57f for example.

kt3k:
Thanks for the kind answer. The example helped me to understand this behavior very well. Thanks!

PR #273, opened 2018-02-22
Grammar tweak

brettcannon:
Add a comma to break up the sentence and clearly separate the two concepts being references, otherwise it reads like its a single concept.


Theodus:
Thanks @brettcannon!

PR #274, opened 2018-02-24
 Let variable has to be assigned immediatly 

gwelr:
From a small test I ran while learning pony, I first though I could do something like: 
```pony

let x: U32
x=4
```
I though this would be correct for as long as the expression `x= ` would only be used once (since the variable is declared using `let`). A quick compilation of my test showed I was wrong. Not only `let` variable could only be assigned once, they also have to be assigned immediatly. This is another difference compared to `var` declared variables.


expressions/variables.md
```diff
@@ -67,9 +75,64 @@ Fields have the same lifetime as the object they're in, rather than being scoped
 
 If the name of a field starts with `_`, it's __private__. That means only the type the field is in can have code that reads or writes that field. Otherwise, the field is __public__ and can be read or written from anywhere.
 
-Just like local variables, fields can be `var` or `let`. They can also have an initial value assigned in their definition, just like local variables, or they can be given their initial value in a constructor.
+Just like local variables, fields can be `var` or `let`. Nevertheless, rules for field assignation differ a bit from variable assignation. No matter the type of the field (either `var` or `let`), either:
```
EpicEric: Use "assignment" instead of "assignation".

Same thing on lines 99, 101, 115.

gwelr: Done, fixed in my latest commit

expressions/variables.md
```diff
@@ -67,9 +75,64 @@ Fields have the same lifetime as the object they're in, rather than being scoped
...
+class Wombat
+  let name: String
+  var _hunger_level: U64
+
+  new ref create(name': String, level: U64) =>
+    name = name'
+    set_hunger_level(level)
+    // Error: field _hunger_level left undefined in constructor
+  
+  fun set_hunger_level(hunger_level: U64) =>
```
Praetonus: Should be `fun ref` since the method modifies the object.

expressions/variables.md
```diff
@@ -67,9 +75,64 @@ Fields have the same lifetime as the object they're in, rather than being scoped
...
+```pony
+class Wombat
+  let name: String
+  var _hunger_level: U64
+
+  new ref create(name': String, level: U64) =>
+    name = name'
+    _hunger_level = level
+
+  fun set_hunger_level(hunger_level: U64) =>
```
Praetonus: Should be `fun ref`.

expressions/variables.md
```diff
@@ -67,9 +75,64 @@ Fields have the same lifetime as the object they're in, rather than being scoped
...
+    _hunger_level = hunger_level // Ok, _hunger_level is of var type
+
+  fun set_name(name' : String) =>
```
Praetonus: Should be `fun ref`.

expressions/variables.md
```diff
@@ -57,6 +57,14 @@ x = 5  // OK
 y = 6  // Error, y is let
 ```
 
+Using `let`instead of `var` also means the variable has to be assigned immediatly.
```
jemc: A few typos here:
- need a space after `let`
- "immediatly" is spelled incorrectly

gwelr:
Typo fixed and `ref` added to `fun` modifying an object. Thank you for the feedback

mfelsche:
Thanks @gwelr !

PR #283, opened 2018-05-03
added section on capability inference and match

sblessing:
This describes is a technique of a non-obvious way to extract iso references using pattern matching.


mfelsche:
Thank you! Nice section.

PR #285, opened 2018-05-12
fix push example in Object Literals

sheerluck:


mfelsche:
Good catch!

Could you also add the missing `use "collections"` to the examples to make them compile?

sheerluck:
I don't think that is needed. Imho we should mention `use "collections"` only once // in the beginning

mfelsche:
Thanks for your contribution!

PR #287, opened 2018-05-16
Made example compile + typo fixed

civ:
- ref example did not compile https://gist.github.com/ponylang-gist/2a7ccf4e3667f06f4a6ad666ee643f26 
- typo in variable name used fixed

### DIFF
--- a/generics/generics-and-reference-capabilities.md
+++ b/generics/generics-and-reference-capabilities.md
@@ -37,7 +37,7 @@ actor Main
     let a = Foo[U32](42)
     env.out.print(a.get().string())
     a.set(21)
-    env.out.print(c.get().string())
+    env.out.print(a.get().string())
 ```
 
 Unfortunately, this doesn't compile. For a generic class to compile it must be compilable for all possible types and reference capabilities that satisfy the constraints in the type parameter. In this case, that's any type with any reference capability. The class works for the specific reference capability of `val` as we saw earlier, but how well does it work for `ref`? Let's expand it and see:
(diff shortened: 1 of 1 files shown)


civ:
maybe add a comment on the added ref at the get method, as val did not need to be there for the get in the previous example?

mfelsche:
Thanks for making this work! :)

Changing the get method to ref is actually not what a `getter` method should need, as it does not change the state of the class and it should not be allowed to.
To make this compile, it should actually be:

```pony
fun get(): this->String ref => _c
```

It should use an Arrow type. This is also introduced further below. I think that actually the example is a bit off here.

civ:
That's why I avoided the arrow type there, as it's explained below like you said... 
I just put the ref there so everything in that constructed ref-example is ref

civ:
`fun get(): String ref => _c.clone()`
works, too, but I don't know why ^^
Is that an option?

mfelsche:
That is indeed an option. 

Another one, which i would prefer, would be to make the return tzpe of `get` a `String box` and add a stance that this needs to be box because this is how the method sees the field `_c`. This would comply with the example, being a simple container.

Or we need to somehow rewrite the narrative of this example, to introduce the arrow type here already.

civ:
I hope I got you right there, I don't understand the capabilities. So is this roughly what you had in mind? 

https://github.com/ponylang/pony-tutorial/compare/master...civ:patch-2?short_path=a319a7b#diff-a319a7b9efaed94c6968b9e97169ccd5

If it is, I'll open a new pull request and we discard this one?
(I put this in a new branch, I don't know how to squash commits, I made these changes in the browser on github.com)

mfelsche:
Closing this in favor of #318 .

Again, sorry for handling this in such an awful way.
Your contribution was very valuable to understanding how to improve this part of the tutorial.
Thank you!

PR #294, opened 2018-05-25
Made example compile + typo fixed

civ:
- ref example did not compile
- typo in variable name used fixed

### DIFF
--- a/generics/generics-and-reference-capabilities.md
+++ b/generics/generics-and-reference-capabilities.md
@@ -37,7 +37,7 @@ actor Main
     let a = Foo[U32](42)
     env.out.print(a.get().string())
     a.set(21)
-    env.out.print(c.get().string())
+    env.out.print(a.get().string())
 ```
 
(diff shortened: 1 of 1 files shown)


civ:
Instead of #287 based on @mfelsche 's feeback there. I hope I got them right ~~

civ:
> Sorry, for stalling this so long!

No worries!

About the changes, I don't mind at all but I would rather not change the narrative myself as I am one of these people who get (thoroughly) confused about reference capabilities ~~

Maybe another option is to keep the code broken and then go along the lines of 
"It doesn't work for ref, we will see how to fix this in a bit [..]" and then comes the iso?
Or yet another, just remove the ref example entirely if it needs the arrow type already, as the point of the example at that position was to show another working class with just the capability swapped (val -> ref)?

mfelsche:
I would prefer to have the `this->ref` example in here, as it shows how to write a getter properly in a capability generic way.

mfelsche:
Closing this in favor of #318 

I have to apologize for the time that passed without me taking care of your contribution.
This really haven't been a good example on how to maintain this open-source project.
Sorry!

PR #296, opened 2018-06-01
Fix typos

tirkarthi:
I have fixed some typos using aspell. The normal diff is not helpful. Please use rich diff in GitHub UI. I have Fixed the following typos : 

progammer => programmer
additonal => additional
assgnment => assignment

You can check changes locally using  below command : 

`git log -p 715de6f8516e034e04a07ec91489c431f614f3b8 --color-words=.`

As a suggestion the rust's tutorial book uses aspell in Travis to catch spelling mistakes. This can be incorporated into the repo. This requires an initial effort of creating a dictionary which are valid with respect to the book but invalid with respect to English like variable names, function names etc and then using the dictionary along with the default dictionary for spell check.

Rust book scripts : https://github.com/rust-lang/book/tree/master/ci

Let me know if you would like a new issue for this so that this can be tracked.

Thanks for the book and Pony.


jemc:
I think adding aspell would be a good idea. Feel free to create a ticket for it if you will.

jemc:
Thanks for the help!

PR #298, opened 2018-06-09
Add note about string literal instances

### LINKED ISSUE #297: Add note about single instance of String literals
As @SeanTAllen pointed out to me, string literals are actually single common instance in the entire program, and not different String instances. Since this could cause some confusion with new programmers (for example, when dealing with identity equality), there should be a note in the Strings literals section to that effect.

EpicEric:
String literals are actually single common instance in the entire program, and not different String instances. This can cause some confusion with new programmers (for example, when dealing with identity equality).

Closes #297.

### DIFF
--- a/expressions/literals.md
+++ b/expressions/literals.md
@@ -4,7 +4,7 @@
 
 Values!
 
-**Where do we want them?***
+***Where do we want them?***
 
 In our Pony programs!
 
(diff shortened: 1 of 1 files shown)


SeanTAllen:
Thanks @EpicEric

PR #304, opened 2018-07-15
Updated object literals section to use correct receivers in the push …

jdhorwitz:
…examples

The code as it was in the Tutorial was not compiling, this change updates it to running code.


aturley:
Thanks @jdhorwitz!

PR #305, opened 2018-07-16
Updated object literals section to use correct receivers in the push …

jdhorwitz:
…examples


EpicEric:
I believe this has already been merged in #304

PR #311, opened 2018-08-07
Fix divide-by-zero code sample

### LINKED ISSUE #309: make code sample in "division by zero" section valid Pony
The following code snipped appears in the "Division by Zero" section:

```pony
let x = 1 / 0
```

This isn't valid Pony because there's no type information. We should figure out a way to make this example into something that (when included as part of a program) will compile.

Either of these two options works for me:

```ponyc
let a: I32 = 1
let b: I32 = 0
let x = a / b
```

or

```ponyc
let x = I32(1) / I32(0)
```

Any other thoughts?

aturley:
The code sample in the section on dividing by zero was not valid Pony
code. It has been fixed and is now valid. This will make it easier for
people who are learning about the language to experiment with the
example.

Fixes #309

### DIFF
--- a/gotchas/divide-by-zero.md
+++ b/gotchas/divide-by-zero.md
@@ -9,7 +9,7 @@ In math, divide by zero is undefined. There is no answer to that question as the
 In Pony, *integer division by zero results in zero*. That's right,
 
 ```pony
-let x = 1 / 0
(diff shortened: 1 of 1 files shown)


SeanTAllen:
@aturley this needs to be rebased.

PR #318, opened 2018-09-05
refactor journey from specific classes to generic class

mfelsche:
by introducing arrow types earlier to make the  example actually work.

@civ this roughly was what i mentioned in #294 and #287 how this section should actually look like.


civ:
I see, thanks! 

Noticed some little typos, can I leave those here to be fixed? Or would you rather have them for a later pull request.

The very first example has the line
> env.out.print(**_c_**.get().string())

Which makes it fail because of the unknown variable c, instead of capabilities / type issues and according error reports. _**c**_ => _**a**_

> Note how applying the **_arraow_** type

 => **_arrow_**

mfelsche:
Thanks, i corrected them. :)

PR #325, opened 2018-10-04
Add explanation of safe-to-write relation between capabilities

Theodus:
This addresses a lack of explanation for when it is safe to write to the fields of objects. Though this relationship can be inferred based on the viewpoint adaptation table, it definitely requires some introduction for newcomers to the language.

### DIFF
--- a/capabilities/combining-capabilities.md
+++ b/capabilities/combining-capabilities.md
@@ -76,3 +76,24 @@ For `trn` and `ref` we need to return a locally immutable reference capability t
 ### Reading from a `tag` variable
 
 This one is easy: `tag` variables are opaque! They can't be read from.
+
+## Writing to the field of an object
+
+Like reading the field of an object, writing to a field depends on the reference capability of the object reference being stored and the reference capability of the origin object containing the field. The reference capability of the object being stored must not violate the guarantees made by the origin object's reference capability. For example, a val object reference can be stored in an iso origin. This is because the tag reference capability guarantees that no alias to that object exists which could violate the guarantees that the iso capability makes.
+
+Here's a simplified version of the table above that shows which reference capabilities can be stored the field of an origin object.
+
+---
+
+| &#x25C1;       | iso object | trn object | ref object | val object | box object | tag object |
+|----------------|------------|------------|------------|------------|------------|------------|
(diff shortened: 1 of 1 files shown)


jemc:
I see nothing to stop us from merging this now, but it's worth noting that this section will change significantly when we implement George Steed's extractive viewpoint adaptation operator in the upcoming RFC (https://github.com/ponylang/ponyc/issues/2815).

capabilities/combining-capabilities.md
```diff
@@ -76,3 +76,24 @@ For `trn` and `ref` we need to return a locally immutable reference capability t
 ### Reading from a `tag` variable
 
 This one is easy: `tag` variables are opaque! They can't be read from.
+
+## Writing to the field of an object
+
+Like reading the field of an object, writing to a field depends on the reference capability of the object reference being stored and the reference capability of the origin object containing the field. The reference capability of the object being stored must not violate the guarantees made by the origin object's reference capability. For example, a val object reference can be stored in an iso origin. This is because the tag reference capability guarantees that no alias to that object exists which could violate the guarantees that the iso capability makes.
+
+Here's a simplified version of the table above that shows which reference capabilities can be stored the field of an origin object.
```
mfelsche: typo: `can be stored **in** the field of an origin object`

capabilities/combining-capabilities.md
```diff
@@ -76,3 +76,24 @@ For `trn` and `ref` we need to return a locally immutable reference capability t
 ### Reading from a `tag` variable
 
 This one is easy: `tag` variables are opaque! They can't be read from.
+
+## Writing to the field of an object
+
+Like reading the field of an object, writing to a field depends on the reference capability of the object reference being stored and the reference capability of the origin object containing the field. The reference capability of the object being stored must not violate the guarantees made by the origin object's reference capability. For example, a val object reference can be stored in an iso origin. This is because the tag reference capability guarantees that no alias to that object exists which could violate the guarantees that the iso capability makes.
```
mfelsche: Is this a typo?

`This is because the tag reference capability guarantees that no alias...` Should this be `val`? If not How do we get from a `val` field on an `iso` to a `tag`?

Theodus: Yes, it should. Thanks!

PR #326, opened 2018-10-11
Clarify that recover wraps a list of expressions

fweimer:
The first example is ambiguous because it uses a type name to invoke the default constructor, so it is unclear to the reader whether an expression or a type follows the recover keyword.

### DIFF
--- a/capabilities/recovering-capabilities.md
+++ b/capabilities/recovering-capabilities.md
@@ -12,11 +12,13 @@ This most straightforward use of `recover` is to get an `iso` that you can pass 
 
 ## What does this look like?
 
+The `recover` expression wraps a list of expressions and is terminated by an `end`, like this:
+
 ```pony
-recover Array[String] end
+recover Array[String].create() end
 ```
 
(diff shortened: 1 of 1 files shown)


capabilities/recovering-capabilities.md
```diff
@@ -12,11 +12,13 @@ This most straightforward use of `recover` is to get an `iso` that you can pass
 
 ## What does this look like?
 
+The `recover` expression wraps a list of expressions and is terminated by an `end`, like this:
+
 ```pony
 recover Array[String] end
```
mfelsche: we could also use `recover Array[String].create() end` here, if that is more clear.

fweimer: True. Is `Array[String]` considered more idiomatic? Then the tutorial should perhaps use it exclusively.

mfelsche: If you write a lot of pony you learn to value the different syntactic sugars, so in this regard i would consider it more idiomatic. On the other side, if this is confusing when teaching about what can be done with `recover`, i think we should use the more explicit variant in this case.

fweimer: @mfelsche Fair enough, I have updated the pull request.

mfelsche:
Thank you!

PR #327, opened 2018-10-15
Method must be ref because it changes  fields

beardhatcode:
Method must be ref because it changes  fields


jemc:
Thanks for the fix!

PR #328, opened 2018-10-28
how-it-works: Mention more languages using type postfix

perlun:
I think it could make sense to mention a few more of these, especially "reasonably popular" languages in various communities.


jemc:
Thanks for contributing!

PR #335, opened 2018-12-11
Update section on auto-recovery to clarify that receiver need not be sendable.

### LINKED ISSUE #321: “Automatic Receiver Recovery” section is incorrect or confusing?
https://tutorial.ponylang.org/capabilities/recovering-capabilities.html#automatic-receiver-recovery

> […] and _that_ means we can call `ref` methods on an `iso` or `trn`, since `iso` and `trn` are both subtypes of `ref`.

> If all the arguments to the method at the call-site are **sendable** […]

> For example, if the method being called wants a `ref` argument, and we pass it an `iso` argument, that's **sendable** at the call-site, so we can still do automatic receiver recovery.

The section seems to imply that _all_ (including the receiver) the arguments must be sendable, but a `trn` is never sendable. Since the receiver is being recovered, I [am thinking](https://github.com/keean/zenscript/issues/32#issuecomment-419689436) it doesn’t need to be sendable. Or if I am incorrect, then how can this work for `trn`?

Note I have never run the Pony compiler in my life. I just started learning by reading the tutorial over the past couple of days. So apologies in advance if this issue is invalid. I don’t have time to sign-up for the mailing list to ask first.

jemc:
This clarification was suggested by @shelby3 [here](https://github.com/ponylang/pony-tutorial/issues/321#issuecomment-444375298).
Resolves #321.

### DIFF
--- a/capabilities/recovering-capabilities.md
+++ b/capabilities/recovering-capabilities.md
@@ -68,7 +68,7 @@ If the `recover` expression could access __non-sendable__ values from the enclos
 
 When you have an `iso` or `trn` receiver, you normally can't call `ref` methods on it. That's because the receiver is also an argument to a method, which means both the method body and the caller have access to the receiver at the same time. And _that_ means we have to alias the receiver when we call a method on it. The alias of an `iso` is a `tag` (which isn't a subtype of `ref`) and the alias of a `trn` is a `box` (also not a subtype of `ref`).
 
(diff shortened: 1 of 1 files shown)


SeanTAllen:
Merging now. Our tutorial setup is busted because of a gitbook issue so, this won't get deployed yet.

PR #347, opened 2019-02-06
[RFC 61] add mod and fld (and variants)

mfelsche:


content/expressions/infix-ops.md
```diff
@@ -88,42 +88,45 @@ The full list of infix operators that are aliases for functions is:
...
-`/?`       | div_partial()? | Partial Division
-`%?`       | rem_partial()? | Partial Remainder
+Operator   | Method         | Description                     | Note
+-----------|----------------|---------------------------------|---------------
+`+`        | add()          | Addition                        |
+`-`        | sub()          | Subtraction                     |
+`*`        | mul()          | Multiplication                  |
+`/`        | div()          | Division                        |
+`%`        | rem()          | Remainder                       |
+`%%`       | mod()          | Modulo                          | Since version `0.27.0`
```
SeanTAllen: Two small things

First,

- I think this should be "Starting with version" as "Since" sounds like its been released.

Second,

We don't have a breaking change on record so the release this goes out in might be `0.26.1`

I think it's reasonable looking at the current open PRs to say `0.26.1` for now.

PR #353, opened 2019-02-11
Add example of async behaviour execution

EpicEric:


SeanTAllen:
`env.out.print` isn't synchronous. `print` is a behavior on an actor.

SeanTAllen:
@EpicEric I think drop the `this` we don't use it elsewhere. Thoughts?

SeanTAllen:
Thanks @EpicEric

PR #354, opened 2019-02-11
Show how to create constructors that vary capabilities

SeanTAllen:


content/capabilities/reference-capabilities.md
```diff
@@ -96,12 +96,37 @@ String box // A string box
...
 ```
 
-__What does it mean when a type doesn't specify a reference capability?__ It means you are using the _default_ reference capability for that type, which is defined along with the type. Here's an example from the standard library:
+__What does it mean when a type doesn't specify a reference capability?__ It means you are using the _default_ reference capability for that type, which is defined along with the type. 
+
+__So do I have to specify a reference capability when I define a type?__ Only if you want to. There are sensible defaults that most types will use. These are `ref` for classes, `val` for primitives (i.e. immutable references) and `tag` for actors.
+
+## How to create objects with different capabilities
+
+When you write a constructor. By default, that constructor will either create a new object with `ref` or `tag` as the capability. In the case of actors, the constructor will always create a `tag. For classes, if defaults to `ref` but you can create with other capabilities. Let's take a look at an example:
```
cquinn: Missing back-tick here.

SeanTAllen: Fixed and pushed.

content/capabilities/reference-capabilities.md
```diff
@@ -96,12 +96,37 @@ String box // A string box
...
+class Foo
+  let x: U32
+
+  new val create(x': U32) =>
+    x = x'
 ```
 
-When we use a `String` we usually mean a string value, so we make `val` the default reference capability for `String`.
+Now when you call `Foo.create(1)`, you'll get a `Foo val` instead of `Foo ref`.
+You could do something like this:
```
cquinn: Maybe say "You could also do something like this:". But it's not clear why you'd do this. Maybe mention again here that this is to provide different caps from the constructor.

SeanTAllen: addressed

cquinn:
Nice.

content/capabilities/reference-capabilities.md
```diff
@@ -96,12 +96,40 @@ String box // A string box
...
 ```
 
-When we use a `String` we usually mean a string value, so we make `val` the default reference capability for `String`.
+When we use a String we usually mean a string value, so we make val the default reference capability for String (but not necessarily for String constructors, see below).
 
 __So do I have to specify a reference capability when I define a type?__ Only if you want to. There are sensible defaults that most types will use. These are `ref` for classes, `val` for primitives (i.e. immutable references) and `tag` for actors.
+
+## How to create objects with different capabilities
+
+When you write a constructor. By default, that constructor will either create a new object with `ref` or `tag` as the capability. In the case of actors, the constructor will always create a `tag`. For classes, if defaults to `ref` but you can create with other capabilities. Let's take a look at an example:
```
jemc: The first period here should be a comma, with the `B` following it correspondingly lowercased.

jemc:
I think this section should also reference how this is distinct from and how it relates to the default capability for the type.

That is, if you define `class val Foo`, that's not enough to construct a `val` - you need `new val` to do that. But if you only have a `new val` constructor, you probably want to make `val` the default capability for `Foo` as well. The interaction between these two different dimensions of capability specifying is part of the confusion, I think, so it's worth noting how they interact.

SeanTAllen:
@jemc I tried writing that but it came out all jumbled. Do you have any thoughts on language for that?

jemc:
On second thought, I think we should leave it out and merge as-is. It's hard to put that in there without getting confusing, I think.

SeanTAllen:
@jemc sounds good. I couldn't figure out how to do it and not be a mess.

PR #355, opened 2019-02-12
Suggest that the reader complete the capabilities chapter before writing them in code.

jemc:
If you try to start putting capabilities into practice before reading the whole chapter, you may get into trouble due to incomplete understanding being filled in by unfounded assumptions. This PR gently suggests this to the reader.


SeanTAllen:
This is good. Should we also point them at the "Learning capabilities section of the website"?

PR #360, opened 2019-02-20
configure Hugo to create a search-index

mfelsche:
This is a requirement for setting up search for the tutorial.

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -20,6 +20,10 @@ uglyURLs = true
   name = "tags"
   weight = 100
 
+[outputs]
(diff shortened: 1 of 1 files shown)


SeanTAllen:
This should include an update to the submodule as well before merging or can that come after?

mfelsche:
This can come after, when merged it just creates an `index.json` file that is not yet used.

SeanTAllen:
I'm merging this despite it breaking RSS index generation as I'm not sure we care.

PR #365, opened 2019-03-01
Add strawman definition of  `digestof` to keywords.md

slfritchie:
After looking through the stdlib code and trying to piece together what the LLVM code that implements `digestof` does, I've created an addition to `keywords.md` for a missing keyword, `digestof`.  I'm not sure that I'm correct, and I'm pretty sure that it's ugly.  Alternative definitions are welcome.

### DIFF
--- a/content/appendices/keywords.md
+++ b/content/appendices/keywords.md
@@ -21,7 +21,7 @@ This listing explains the usage of every Pony keyword.
 | compile_error | will provoke a compile error 
 | continue | continues a loop with the next iteration
 | consume | move a value to a new variable, leaving the original variable empty
-| digestof | create a `USize` value that represents the identity of the original value
+| digestof | create a `USize` value that summarizes the Pony object, similar to a Java object's `hashCode()` value.
 | do | loop statement, or after a with statement
 | else | conditional statement in if, for, while, repeat, try (as a catch block), match 
 | elseif | conditional statement, also used with ifdef
--- a/content/appendices/keywords.md
+++ b/content/appendices/keywords.md
@@ -21,6 +21,7 @@ This listing explains the usage of every Pony keyword.
 | compile_error | will provoke a compile error 
 | continue | continues a loop with the next iteration
 | consume | move a value to a new variable, leaving the original variable empty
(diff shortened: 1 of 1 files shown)


jemc:
This operator used to be called `identityof` but it was determined that "identity" was misleading when in some cases two different values could have the same resulting "identity". For example, on 32-bit systems, USize is 32-bits, and we can't possibly represent all U64 values with unique 32-bit identities.

So we have a deterministic result that will usually be unique, but is not guaranteed to be unique in all cases.

slfritchie:
A 64-bit platform would have the same collision problem for bigger-than-64 bit 128-bit primitives.

I suppose it's also possible to have a collision between a `USize` and a not-machine-native Pony object pointer/tag/thingie where the former coincidentally has the same value of the latter on a 64-bit CPU, right?  If you care about the difference in that case, then you want to use `is` instead?

I'd never used `digestof` until @jtfmumm suggested using it for some printf-style debugging.  I've always used `@printf("blah 0x%lx blah blah\n".cstring(), pony_object)`, but that trick only works when using the FFI.

Is it safe to say that `digestof` has most/all of the power and limitations of Java's `hashCode`?  If so, perhaps we could borrow a pithy summary from the Java community?

jemc:
> A 64-bit platform would have the same collision problem for bigger-than-64 bit 128-bit primitives.

Yeah, definitely - I was just trying to give one easy (and common) example of collision.

> I suppose it's also possible to have a collision between a `USize` and a not-machine-native Pony object pointer/tag/thingie where the former coincidentally has the same value of the latter on a 64-bit CPU, right?

Yeah, this too.

> If you care about the difference in that case, then you want to use `is` instead?

Yeah, exactly - that's what `is` is for - it takes type and bit width into account to have no false positives / collisions. The `digestof` method is only really useful when you want a comparison value you can do math/bucketing on for hash purposes, and you can deal with the possibility of false positives by having "another line of defense" behind the hash comparison.

That's why it was renamed from `identityof` to `digestof` - it's really just about supporting *hashing* of identities rather than identities themselves.

PR #369, opened 2019-03-05
Add reference capability matrix section

### LINKED ISSUE #331: Explain reference capability matrix in the tutorial
TL;DR: I believe we should add a reference capability matrix section in a friendly way to help people unfamiliar with actor/capability systems get a better intuition on how they work.

----

One resource that I've personally found very useful when learning about Pony capabilities is the reference capability matrix, reproduced below from CLEBSCH et al. 2015:

-- | Deny global read/write aliases | Deny global write aliases | Allow all global aliases
----- | ----- | ----- | -----
**Deny local read/write aliases** | *iso*
**Deny local write aliases** | trn | *val*
**Allow all local aliases** | ref | box | *tag*
-- | (Mutable) | (Immutable) | (Opaque)

This table also appears in other papers, and talks about Pony, and it has some benefits as it helps visualizing some operations and ideas, such as:

- Sendable capabilities (main diagonal)
- Ephemeral subtyping (walking down/to the right)
- Capability recovery (walking up)
- Impossible capabilities (top-right of the matrix is empty)
- etc.

Currently, the tutorial [explicitly enumerates](https://tutorial.ponylang.io/capabilities/reference-capabilities.html) the capabilities based on relevance, and listing individual attributes, without featuring the matrix anywhere. On one hand, you can get a firm grasp of what each capability does on its own, but I find this approach confusing as it's not very clear how each capability relates to one another. In the next sections, the tutorial says how they relate to one another, but a reader -- such as myself -- might still struggle to grasp the idea of [viewpoint adaptation](https://tutorial.ponylang.io/capabilities/combining-capabilities.html) without a solid base.

However, I also believe that the reference capability matrix above can be specially confusing to new users that haven't used actor languages before, as "local" and "global" in terms of references are not clear concepts for people coming from other programming backgrounds.

I've opened this issue to get feedback on how we could add the same visual intuition that the reference capability matrix provides for people that have never worked before with an actor system, or capabilities, or parallel programming, or all three.

EpicEric:
Closes #331

### DIFF
--- a/content/reference-capabilities/capability-matrix.md
+++ b/content/reference-capabilities/capability-matrix.md
@@ -1,0 +1,55 @@
+---
+title: "Reference Capability Matrix"
+section: "Reference Capabilities"
+menu:
+  toc:
+    parent: "reference-capabilities"
+    weight: 110
+toc: true
+---
+
+At this point, it's quite possible that you read the previous sections in this chapter and are still pretty confused about the relation between reference capabilities. It's okay! We have all struggled when learning this part of Pony, too. Once you start working on Pony code, you'll get a better intuition with them.
+
+In the meantime, if you still feel like all these tidbits in the chapter are still scrambled in your head, there is one resource often presented with Pony that can give you a more visual representation: the __reference capability matrix__.
+
+It is also the origin of the concept behind each capability in Pony, in the sense of how each capability denies certain properties to its reference -- in other words, which guarantees a capability makes. We will explain what that actually means before presenting the matrix.
+
+## Local and global aliases
+
+Before anything else, we want to clarify what we mean by "local" and "global" aliases.
+
+A local alias is a reference to the same variable that exists in the same actor. Whenever you pass a value around, and it's not the argument of an actor's behavior, this is the kind of alias we are working with.
+
+On the other hand, a global alias is a reference to the same variable that can exist in a _different_ actor. That is, it describes the properties of how two or more actors could interact with the same reference.
+
+Each reference capability in Pony is actually a pair of local guarantees and global guarantees. For instance, `ref` doesn't deny any read/write capabilities inside the actor, but denies other actors from reading or writing to that reference.
+
+You may recall from the [Reference Capability Guarantees](guarantees.md) section that mutable references cannot be safely shared between actors, while immutable references can be read by multiple actors. In general, global properties are always as restrictive or more restrictive than the local properties to that reference - what is denied globally must also be denied locally. For example, it's not possible to write to an immutable reference in either a global or local alias. It's also not possible to read from or write to an opaque reference, `tag`. Therefore, some combinations of local and global aliases are impossible, and have no designated capabilities.
+
+## Reference capability matrix
+
+Without further ado, here's the reference capability matrix:
+
+---
+
+&nbsp; | Deny global read/write aliases | Deny global write aliases | Don't deny any global aliases
+----- | ----- | ----- | -----
+__Deny local read/write aliases__ | __`iso`__ | |
+__Deny local write aliases__ | `trn` | __`val`__ |
+__Don't deny any local aliases__ | `ref` | `box` | __`tag`__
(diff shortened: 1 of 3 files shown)


EpicEric:
I feel that this section got messier than I had planned. I still want to revisit the text before I feel comfortable merging this.

content/reference-capabilities/capability-matrix.md
```diff
@@ -0,0 +1,53 @@
...
+* __Aliasing__. This is a bit less intuitive than the other points, since it only applies to `iso!` (which aliases to `tag`), and `trn!` (which aliases to `box`). If you picture a mirror in the anti-diagonal of the matrix (on top of `ref` and `val`), we can imagine that `iso` would be reflected on top of `tag` and `trn` would be reflected on top of `box`. Why? Well, our guarantees must be respected in new aliases to our references, . `iso` forbids reads/writes anywhere, so the new alias must prevent the programmer from doing any of these two actions locally and globally, which is why we end up with `tag`; and `trn` denies local and global writes (rendering it immutable), _but_ allows local reads, which gives us the locally immutable `box`, which is another way to say "it cannot be written to, at least from this reference". And since all other capabilities alias to themselves, we can simply follow the same rule for ephemeral subtyping: walking downwards and/or to-the-right.
```
EpicEric: I'm not particularly fond of this item, since it doesn't actually give much visual intuition, and might drop it altogether.

SeanTAllen:
Ping @jemc and I when you need a final review of this.

EpicEric:
Still might need adjustments, but it is ready for review.

@SeanTAllen @jemc

content/reference-capabilities/capability-matrix.md
```diff
@@ -0,0 +1,55 @@
...
+
+Before anything else, we want to clarify what we mean by "local" and "global" aliases.
+
+A local alias is a reference to the same variable that exists in the same actor. Whenever you pass a value around, and it's not the argument of an actor's behavior, this is the kind of alias we are working with.
+
+On the other hand, a global alias is a reference to the same variable that can exist in a _different_ actor. That is, it describes the properties of how two or more actors could interact with the same reference.
+
+Each reference capability in Pony is actually a pair of local guarantees and global guarantees. For instance, `ref` doesn't deny any read/write capabilities inside the actor, but denies other actors from reading or writing to that reference.
+
+You may recall from the [Reference Capability Guarantees](guarantees.md) section that mutable references cannot be safely shared between actors, while immutable references can be read by multiple actors. In other words, this implies that global properties are as restrictive or more restrictive than the local properties to that reference; it's not possible to write to an immutable reference in a local alias, for example.
```
jemc: This paragraph still seems a bit scattered to me. I've tried to suggest some changes that make it a bit more digestible, in my opinion:

```suggestion
You may recall from the [Reference Capability Guarantees](guarantees.md) section that mutable references cannot be safely shared between actors, while immutable references can be read by multiple actors. In general, global properties are always as restrictive or more restrictive than the local properties to that reference - what is denied globally must also be denied locally. For example, it's not possible to write to an immutable reference in either a global or local alias.
```

EpicEric: I have added your change, but I'll review this paragraph myself later.

EpicEric:
@jemc I've added a few sentences to the paragraph you've mentioned, to try and make its conclusion more clear.

I have also squashed and rebased the branch.

jemc:
> I have also squashed and rebased the branch.

But now how will I know what the new sentences are? I hate squashing during review :disappointed_relieved: - in the future we can just squash on merge.

Can you point out what the changes were so I don't have to hunt and peck for them?

EpicEric:
I simply included two sentences to the end of your suggestion:

```diff
- [snip] For example, it's not possible to write to an immutable reference in either a global or local alias.
+ [snip] For example, it's not possible to write to an immutable reference in either a global or local alias. It's also not possible to read from or write to an opaque reference, `tag`. Therefore, some combinations of local and global aliases are impossible, and have no designated capabilities.
```

PR #373, opened 2019-03-29
Reorder "Reference Capabilities" sections

EpicEric:
The "Combining Capabilities" section of the tutorial covers topics that are more advanced than the previous ones. Placing it after "Capability Subtyping" means that someone reading the tutorial could start writing toy programs that send simple messages to other actors, such as `String iso`, right after reading "Passing and Sharing References".


SeanTAllen:
@jemc are you good with this?

jemc:
Yes! It's a great idea.

PR #375, opened 2019-04-20
Avoid "give out" which sounds like sharing with other actors

schnittchen:


SeanTAllen:
Nice. Thanks @schnittchen. Awesome first contribution.

Issue #385, opened 2019-05-30
Type parameter type defaults seem to be missing


schnittchen:
I'm referring to things like the `[B: ... ref = Array[A!]]` part here https://stdlib.ponylang.io/src/itertools/iter/#L259.

I could not find this in the tutorial.

Please feel free to correct me on the terminology.

rhagenson:
As this issue sits at the intersection of Pony generics it is a little more complicated, but not unreasonable for a first issue. Anyone wishing to tackle this issue, feel free to reach out to me. First we will need to decide where in the Tutorial is best for this and a minimal example to show type parameterization with a default.

EpicEric:
Closed in #467.

PR #386, opened 2019-05-31
`=` also used for defaults (re #385)

schnittchen:


content/appendices/symbol-lookup-cheatsheet.md
```diff
@@ -52,6 +52,7 @@ Here is a more elaborate explanation of Pony's use of special characters: (a lin
 | `;`  | only used to separate expressions on the same line |
 | `=`  | (1) (destructive) assignment |
 |    | (2) in: use alias = packname |
+|    | (3) default arguments and default types |
```
SeanTAllen: I think this should be two entries. Each should get you to what you need with a search:

So something like

`Supply default type for generics`
`Supply default argument for method`

schnittchen: I'll take that, reversing the order

SeanTAllen:
Thanks @schnittchen!

PR #400, opened 2019-09-10
Capitalize non-code mentions of Pony

RHagenson:
As Pony is the name of the language, any plaintext/non-code mentions of it should have the name capitalized.

Note that the String Literals that mention `pony` as a variable name have not been changed as these were the only code mentions of pony


EpicEric:
Thanks!

PR #407, opened 2019-10-21
Add Elixir to list of actor model languages

batate:
I added Elixir to the list of languages supporting actors because it's probably the most likely actor model that a Ruby programmer is likely to encounter, and that's a pretty big ecosystem.


SeanTAllen:
Thanks @batate

PR #411, opened 2019-11-14
Improve example that illustrates what "as" expressions are for

philix:
The previous example has an explicit Array type annotation and the
example containing the "as" expression does not need that since it wants
to demonstrate how the coercion can be achieved by expressing the type
of the elements in the Array after with "as".


SeanTAllen:
@jemc I'm neutral.

jemc:
That's good enough for me. :grin: 

Thanks for your contribution!

PR #412, opened 2019-11-19
Resolve #336; is/isnt keywords use for identity comparison

### LINKED ISSUE #336: Keyword listing is missing `isnt`
https://tutorial.ponylang.io/appendices/keywords.html doesn't have `isnt`. 

😢 

it should then we can be

😄

rhagenson:
Added third use of `is` keyword for identity comparison and `isnt` entry as negative identity comparison in keyword table.


SeanTAllen:
Awesome. Thank you.

PR #414, opened 2019-11-19
Resolve #413; add not as keyword

### LINKED ISSUE #413: Add not as keyword
Noticed while adding `isnt` to keyword listing (#336) that `not` is not listed as a keyword.

rhagenson:
Added `not` as keyword for logical negation.


SeanTAllen:
Thanks. FYI. the general preference is to have resolves #BLAH etc to be in the commit body, not the "header".

rhagenson:
Understood on the resolves #BLAH in body not header.

PR #417, opened 2019-11-19
Add self-referential constructor call

### LINKED ISSUE #284: Tutorial lacks mention of syntax for constructor style: let x: T = x.create()
Today in the #ponylang IRC channel, `brainproxy` asked a question about this syntax in the Pony Patterns doc, in the "Access" section.

    let _data: collections.Map[String, I64] = _data.create()

Neither `brainproxy` nor I can find an example or explanation in the Pony tutorial that mentions this style use.  By my own skim, I found one almost-but-not-quite in the "Methods" section of the tutorial, the "Calling" subsection:

    var b: Foo = a.from_int(3)

rhagenson:
Fixes #284 by adding the `var a: T => a.create()` syntax after introduction of constructors.

### DIFF
--- a/content/expressions/methods.md
+++ b/content/expressions/methods.md
@@ -133,6 +133,16 @@ class Bar
     var b: Foo = a.from_int(3)
 ```
 
+We can even reuse the variable name in the assignment expression to call the constructor.
+
+```pony
+class Bar
+  fun f() =>
+    var a: Foo = a.create()
+```
+
+Here we specify that `var a` is type `Foo`, then proceed to use `a` to call the constructor, `create()`, for objects of type `Foo`.
+
 ## Default arguments
 
 When defining a method you can provide default values for any of the arguments. The caller then has the choice to use the values you have provided or to provide their own. Default argument values are specified with a `=` after the parameter name.


rhagenson:
Don't accept this yet. I realized I put `.create()` in parentheses. One minute.

rhagenson:
There we are, oversight on my part that obviously "(create())" reads terribly.

content/expressions/methods.md
```diff
@@ -115,7 +115,7 @@ class Bar
 
 ```
 
-Constructors can also be called on an expression. Here an object is created of the same type as the specified expression - this is equivalent to directly specifying the type.
+Constructors can also be called on an expression. Here an object is created of the same type as the specified expression -- this is equivalent to directly specifying the type.
```
EpicEric: Could you revert the hyphen addition? As it isn't related to the issue.

rhagenson: Absolutely. Force of habit perpetually fixing such things in my own work. Will get the change made later today.

rhagenson: @EpicEric the hyphen is now removed

EpicEric:
Thanks, @rhagenson.

PR #436, opened 2020-06-26
Clarifying when a lambda requires a `ref` reference capability

bradleyscollins:
A sentence in the paragraph that explains when a lambda requires `ref` reads awkwardly. The biggest problem is that it appears to be missing a comma:

> In a lambda that **captures references, this needs** to be `ref` if the function needs to modify any of the captured variables or call `ref` methods on them.


jemc:
Thanks!

PR #437, opened 2020-06-28
Split talk about 's and numbers to avoid prime confusion

grahamc:
This paragraph talking about numbers lead me to a few minutes wondering why this rule about prime numbers exists, before I realized you meant prime-the-single-quote.


jemc:
Thanks!

PR #438, opened 2020-06-28
Type aliases: Restore personhood to people without a home

grahamc:


SeanTAllen:
To avoid any potential issues in the future, I think that switching from person to something like animal classification would be a better example here.

https://a-z-animals.com/reference/animal-classification/

Would you be up for making the slightly larger change to stop using Person here and switch to Animal and Classification qualities like Genus and Species?

SeanTAllen:
Thanks @grahamc.

https://github.com/ponylang/pony-tutorial/issues/446 has been created to further work on this issue.

PR #441, opened 2020-07-07
Clear up confusion about statements as expressions

### LINKED ISSUE #440: Statement v. expressions interchanging
I noticed this interchanging while reading [Everything is an expression](https://tutorial.ponylang.io/expressions/control-structures.html#everything-is-an-expression) where the following is said:

> In Pony there are no statements, there are only expressions; everything hands back a value. Your `if` **statement** hands you back a value. Your `for` loop (which we’ll get to a bit later) hands you back a value.

Above emphasis is mine. Certainly this specific instance must be changed, but we make similar expression/statement swaps in the [Keywords](https://tutorial.ponylang.io/appendices/keywords.html), [Use statement](https://tutorial.ponylang.io/packages/use-statement.html), [Match expression](https://tutorial.ponylang.io/pattern-matching/match.html) (where it is later called a "match statement" one before the "Matching on values" section), [Variables](https://tutorial.ponylang.io/expressions/variables.html), and so on. 

It seems there are more instances of "statement" rather than "expression" so before making changes I wanted to open up the conversation on such a sweeping change.

Related point, if `if`, `match`, `for`, variable assignment (asides on [destructive read](https://tutorial.ponylang.io/types/classes.html) talks above this) and so on are expressions, are there instances of Pony statements which will then need to be specifically explained as being unlike the rest of Pony?

jemc:
Issue #440 was discussed in today's sync call. I mentioned there that I think it's least confusing to just avoid saying "these are not statements" while still saying positively that "these are all expressions".

Resolves #440.

### DIFF
--- a/content/expressions/control-structures.md
+++ b/content/expressions/control-structures.md
@@ -70,7 +70,7 @@ else
 ```
 Here it is not obvious whether the `else` is an alternative to the first or the second `if`. In fact here the `else` relates to the `if(b)` so our example contains a bug. Pony avoids this type of bug by handling `if` and `else` differently and the need for `elseif` comes out of that.
 
-## Everything is an expression
+## Control structures are expressions
 
 The big difference for control structures between Pony and other languages is that in Pony everything is an expression. In languages like C++ and Java `if` is a statement, not an expression. This means that you can't have an `if` inside an expression, there has to be a separate conditional operator '?'.
 
--- a/content/expressions/control-structures.md
+++ b/content/expressions/control-structures.md
@@ -74,7 +74,7 @@ Here it is not obvious whether the `else` is an alternative to the first or the 
 
 The big difference for control structures between Pony and other languages is that in Pony everything is an expression. In languages like C++ and Java `if` is a statement, not an expression. This means that you can't have an `if` inside an expression, there has to be a separate conditional operator '?'.
 
-In Pony there are no statements, there are only expressions; everything hands back a value. Your `if` statement hands you back a value. Your `for` loop (which we'll get to a bit later) hands you back a value.
+In Pony control flow statements like this are all expressions that hand back a value. Your `if` statement hands you back a value. Your `for` loop (which we'll get to a bit later) hands you back a value.
 
 This means you can use `if` directly in a calculation:
 


EpicEric:
Shouldn't the section title "Everything is an expression" also be changed?

content/expressions/control-structures.md
```diff
@@ -74,7 +74,7 @@ Here it is not obvious whether the `else` is an alternative to the first or the
 
 The big difference for control structures between Pony and other languages is that in Pony everything is an expression. In languages like C++ and Java `if` is a statement, not an expression. This means that you can't have an `if` inside an expression, there has to be a separate conditional operator '?'.
 
-In Pony there are no statements, there are only expressions; everything hands back a value. Your `if` statement hands you back a value. Your `for` loop (which we'll get to a bit later) hands you back a value.
+In Pony control flow statements like this are all expressions that hand back a value. Your `if` statement hands you back a value. Your `for` loop (which we'll get to a bit later) hands you back a value.
```
EpicEric, line 77: I'd probably include `while` loops in here, if the intent is to list statements which usually aren't expressions in other programming languages.

rhagenson:
The issue has not been resolved in my opinion. This PR addresses the immediate point of clarifying the language of "Everything is an expression" but it does not address the more widely reaching intent of the issue which was consistent clarity on the point of statement versus expression.

Sean had my thinking correct on Sync yesterday in that the issue was written from the point-of-view that (paraphrasing) "statements do not return values, expressions return values". I see no issue with the idea that expressions can be statements, but want clarity in the tutorial so that a new user can differentiate, for example, `if` as a statement and `if` as an expression. The only use case I can think of in my head for when this would matter during learning is when using an `if` block as the last expression in a function block and any resulting confusion on why that function is returning `None` (the implicit `else` return).

```pony
class A
  var value: I8

  new create() =>
    value = 0
  fun ref add1() =>
    value = value + 1
  fun get(): I8 =>
    if value == -1 then 
      value
    end
```

Playground link which shows the error in this code: https://playground.ponylang.io/?gist=a6302e798f6940cf8051c3160a59ae50

As a newcomer, I do not think it would be clear that the last expression/return here is the `if` block, not the `value` inside the `if` block.

Any rebuttals to having a clear expression/statement terminology along the lines of return value/no return value?

jemc:
Basically, my thinking on this is to suggest that "all expressions are statements, but not all statements are expressions", with expressions being the subclass of statements that do have a result value.

I'm happy to retract this PR if you want to address this part of a text in a different way.

rhagenson:
I take no issue with merging this PR as long as there is an understanding that I will review the expression/statement terminology usage across the Tutorial to meet our agreed upon understanding of their relation to one another being that one is a subclass of the other.

jemc:
Sure, we can iterate as needed with whatever improvements you see fit.

SeanTAllen:
We are going to merge this, Ryan if there are specifics you would like to improve further, please create an issue so we don't end up losing track of them.

rhagenson:
Certainly can do. That was my intention if I reach another larger conversation point; otherwise I was going to make smaller edits for the other places I mentioned in the original issue to reflect our shared understanding of statement versus expression in the tutorial.

PR #442, opened 2020-07-10
Viewpoint adaptation changes incl. `trn->trn = box`

jasoncarr0:
Here's my suggestions for an update to the viewpoint adaptation page, which includes a small update to wording at the beginning, and a refreshed `trn`origin section, which explains why `trn->trn` has to be `box` for soundness in the current version.


jasoncarr0:
Going to re-open two new PRs instead

PR #443, opened 2020-07-10
Combining capabilities: Rework the "see its own fields" wording of viewpoint adaptation.

jasoncarr0:
Explanation below:

While it is true that access to receivers and fields will follow
this, in that `this` is just a normal variable, I feel this outlook
could be confusing when understood with receiver recovery, and
changing of capabilities.

A reader might conclude or speculate that a variable with type iso,
can only see its own contents with those viewpoints, but the notion
of an object accessing itself is most closely aligned with calling
a function (or behavior). Since this will almost certainly be
`fun ref` and `fun box`, the reader may become confused by
reading it as though it would apply to those functions on `iso`
variables.

I'm not implying that it would be sensible to conclude that, but
that a reader may get confused that that would seem to be implied.

### DIFF
--- a/content/reference-capabilities/combining-capabilities.md
+++ b/content/reference-capabilities/combining-capabilities.md
@@ -14,9 +14,9 @@ This is because all the guarantees that the __origin__ reference capability make
 
 # Viewpoint adaptation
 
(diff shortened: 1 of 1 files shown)


SeanTAllen:
@jemc can you review this one?

Issue #439, opened 2020-07-01
Document that partial application doesnt work with iso and trn parameters


SeanTAllen:
As seen via this example where it gets a compilation error:

https://playground.ponylang.io/?gist=056f09e2ab1be95390b103ad37833e3f

You can't use partial application with iso/trn parameters in the current implementation. Partial application is sugar over object literal creation.

@jemc states:

> I think it's an unfortunate limitation based on the fact that partial application is compiled to an object literal with captures, and the compiler doesn't know that it's a "use once" object literal
>
> In order to treat it as a "use once" object literal, it would need to be an iso object literal, but that has viewpoint adaptation implications that may mess up other arguments you're trying to send, so it can't be done that way in the general case

rhagenson:
Closed by #445

anacrolix:
Thanks, but could you provide a workaround for the sample code too? Now it's documented that it can't be done the way I originally expected, how should I do it?

jemc:
@anacrolix - I copied the original playground code from the OP in this ticket (the example with `Herp` and `Derp`), and adapted it to use a class called `CreateHerp` rather than partial application syntax. The purpose of doing this is so that I could make the final application a `fun iso` that calls `consume this`, to represent to the compiler that this partial application object is one-time-use-only, thus upholding the required uniqueness guarantees that were missing in the version that uses partial application.

This may not be sufficient for every case, but it at least works for the case in the playground example. Here's the sample code: https://playground.ponylang.io/?gist=056f09e2ab1be95390b103ad37833e3f

```pony
class iso Derp

actor Herp
  new create(derp: Derp, env: Env) =>
    env.out.print("hooray!")

class iso CreateHerp
  let _derp: Derp
  new iso create(derp: Derp) => _derp = consume derp
  fun iso apply(env: Env): Herp => Herp((consume this)._derp, env)

actor Main
  new create(env: Env) =>
    let partially_applied_thing = CreateHerp(Derp)
```

anacrolix:
Thanks!

PR #445, opened 2020-07-10
Rework the partial application object literal section (fixes https://github.com/ponylang/pony-tutorial/issues/439)

jasoncarr0:
Reworked the "Partial application is just an object literal" section.

It's hard not to combine multiple changes here, but this was an opportunity to make the admittedly bold change that we should use `lambda` here as it's likely far more familiar than `object literal`, and it reads better in the title imo.

Besides that, just re-order before the section that references it, change 'captures' to 'captures aliases of', and add an explaining sentence why we can't consume unique references, fixing #439

### DIFF
--- a/content/expressions/partial-application.md
+++ b/content/expressions/partial-application.md
@@ -55,6 +55,12 @@ f(3)
 
 Here, we bound the `mul` argument but left `add` unbound.
 
+## Partial application is just a lambda
+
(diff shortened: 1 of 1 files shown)


SeanTAllen:
@rhagenson can you review this one?

rhagenson:
@SeanTAllen LGTM. 

I think this makes it clearer why partial application can/cannot be used in certain ways. Lambdas are currently covered in the preceding section (Object Literals) so are by all means are reasonably familiar to a reader.

SeanTAllen:
Thanks @jasoncarr0

PR #450, opened 2020-11-19
Fix typo

Yam76:
"it's" should be "its"


SeanTAllen:
Thanks!

rhagenson (approved):
Great catch. Thank you for the fix, @Yam76!

PR #461, opened 2021-02-02
update control-structure.md to clarify what is needed to avoid using …

carlomunguia:
…None in control structures, per Sean Allen / feedback from pony-zulip

### DIFF
--- a/content/expressions/control-structures.md
+++ b/content/expressions/control-structures.md
@@ -122,6 +122,20 @@ actor Main
 
 This will give __x__ the value "Sarah" as it is the last name in our list. If our loop has 0 iterations, then the value of its `else` block will be the value of __x__. Or if there is no `else` block, the value will be `None`.
 
+You can also avoid needing `None` at all by providing a __default value__ for when the loop has __0 iterations__ by providing an `else` block.
+
+```pony
+  actor Main
+  new create(env: Env) =>
+    var x: String =
+      for name in Array[String].values() do
+        name
+      else
+        "no names!"
+      end
+    env.out.print("x is " + s)
+```
+
 ```pony
 actor Main
   new create(env: Env) =>
--- a/content/expressions/control-structures.md
+++ b/content/expressions/control-structures.md
@@ -153,7 +153,6 @@ actor Main
 
 And finally, here the value of __x__ is "no names!"
 
-
 ## Loops
 
 `if` allows you to choose what to do, but in order to do something more than once, you want a loop.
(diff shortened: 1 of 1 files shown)


content/expressions/control-structures.md
```diff
@@ -122,6 +122,20 @@ actor Main
...
+  actor Main
+  new create(env: Env) =>
+    var x: String =
+      for name in Array[String].values() do
+        name
+      else
+        "no names!"
+      end
+    env.out.print("x is " + s)
+```
```
rhagenson, line 137: You added the new code block but did not remove the old code block. Will you make the edits so that your additions are incorporated with the example flow listed below?

That is the examples on this page proceed as:
1. Showing an if-else with return type `(String | Bool)`
2. Showing an if-else with return type `(String | None)`
3. Loop with literal Array[String] and no else block with return type `(String | None)`
4. Loop with literal (empty) Array[String] and no else block with return type `(String | None)`
5. Your loop with literal (empty) Array[String] and default else block with return type `String`

Some of the intervening text between examples will need to be rewritten to allow this flow -- for example, it would not longer make sense so say "And lastly, here x would be None." if that example (4 above) precedes your additional example (5 above).

content/expressions/control-structures.md
```diff
@@ -122,6 +122,20 @@ actor Main
 
 This will give __x__ the value "Sarah" as it is the last name in our list. If our loop has 0 iterations, then the value of its `else` block will be the value of __x__. Or if there is no `else` block, the value will be `None`.
 
+You can also avoid needing `None` at all by providing a __default value__ for when the loop has __0 iterations__ by providing an `else` block.
+
+```pony
+  actor Main
```
rhagenson, line 128: Although ponyc is not particular about it, please dedent this `actor Main` so there are not preceding spaces/tabs.

content/expressions/control-structures.md
rhagenson: Not sure I was clear on this following proofreading the above comment, but I intend to mean "your loop" should replace the old problematic loop example as well as reorganize the examples to follow the above example flow. Your added code block should be followed up by the "Here the value of x is "no names!"" intervening text.

rhagenson (changes requested):
Carlo, thank you for submitting these changes! I hope my requested edits make sense. Please let me know if they do not. 

My requested edits are with respect to incrementally deepening the reader's understanding on expression return values.

rhagenson:
@carlomunguia Please let me know when this is ready to review again. It looks like the latest was a force push, but that did not include the reorganization of examples.

rhagenson (approved):
Looks like every requested change has been made. Thank you for your contribution!

PR #462, opened 2021-02-03
Correct type of x in example

SilvanCodes:
The example of loop iteration with an "else" branch that is also a `String` now correctly specifies the type of `x` as `String`. The following `match` statement is thereby obsolete.


rhagenson:
This is being working on in #461 by Carlo with further changes to the flow of examples. With this change to drop `None` I decided a rearrangement would better show Pony's expression return typing.

rhagenson:
Closing as #461 made the necessary changes.

SeanTAllen:
Thanks for your work @SilvanCodes even if it didn't get merged as someone else jump on it.

PR #467, opened 2021-04-12
Type parameter defaulting

schnittchen:


rhagenson:
@schnittchen Thank you for opening this PR! I apologize it took me a couple days to review it.

Looking at the issue you raised (#385), it appears the reported problem at that time was no presentation of defining a generic with default parameter. Such as `[B: ... ref = Array[A!]]` below:

```pony
fun ref collect[B: Seq[A!] ref = Array[A!]](coll: B): B^ =>
```

This PR does not show a working generic with default type parameter. Would you mind updating your PR so there is, for example, a `Bar` that mirrors `Foo` which applies a default type parameter then shows how use of that generic without the type parameter is possible -- as long as the function argument is the same type as the default parameter?

schnittchen:
I've added a complete example. I've also added a small change at the top of the page, to make it clear that the ref capability is part of the type parameter.

Should be squashed.

docs/generics/index.md
```diff
@@ -82,6 +83,31 @@ class FooString
...
+    _c = c
+
+  fun get(): A => _c
+
+  fun ref set(c: A) => _c = c
+```
+
+Now, when the default type parameter is the desired one, it can simply be omitted. But it is still possible to be explicit or use a different type.
+
+```
```
EpicEric, line 105:
```suggestion
```pony
```

In order to lint successfully.

EpicEric (changes requested):

schnittchen:
@EpicEric was so quick with his suggestion, one could think he's a bot

docs/generics/index.md
```diff
@@ -82,6 +83,31 @@ class FooString
   fun ref set(c: String val) => _c = c
 ```
 
+### Type parameter defaults
+
+Sometimes the same parameter type is used over and over again, and it is tedious to always specify it when using the generic class.
+The class `Bar` expects its type parameter to be a `val USize` by default:
```
rhagenson, line 89: This should state `USize val` as the refcap is written after the type.

docs/generics/index.md
```diff
@@ -82,6 +83,31 @@ class FooString
...
+
+```pony
+class Bar[A: Any val = USize]
```
rhagenson, line 92: Although it is implied by the default refcap of a `USize` will you change this to match how you wrote it above as `USize val`?

rhagenson (changes requested):
Thank you for adding this more extensive explanation of the use of type parameter defaults. May I request a few minor changes to make this explanation slightly easier for a new Pony programmer to understand on first read through?

docs/generics/index.md
schnittchen: I wasn't aware that the RHS is subject to "reference capability defaulting" again. So `[A: Any val = USize val]`?

rhagenson: I don't think the RHS is subject to defaulting like my previous comment suggested. After looking at generic parameters in the stdlib and reconsidering this minor change I have reversed my opinion to this change NOT being needed.

The reason being that this style of the default refcap being repeated so `Any **val** = USize **val**` is not the norm in the stdlib. As well, I think the repetition, although superficially mirroring the above change, is distracting since `Any **val**` is already specified.

schnittchen: This leads to a interesting question:

Is this legal, the default having a different ref cap? (I changed it so the default is valid, "val" is acceptable where a "box" is expected)?

```
class Bar[A: Any box = USize val]
```

If yes, it should probably be mentioned, including that `[A: Any box = USize]` means `[A: Any box = USize box]` and that it is usual to leave the second ref cap out.
If not, I would just keep this part as it is.

rhagenson: You should run some tests because I do not know without looking if this is legal and works the way you mention. If it does work that way, I agree that adding additional information here would be useful.

schnittchen: `class Bar[A: Any box = USize val]` is legal. I'll make a proposal in a separate commit.

ergl: > including that [A: Any box = USize] means [A: Any box = USize box]

I don't think that's correct, it will still mean `USize val`. See this example on how `[A: Any tag = USize]` will fail to compile:

https://playground.ponylang.io/?gist=576c7950021c592647febc8016120e03

schnittchen: Thank you @ergl. I've updated accordingly, hope I got it right this time.

ergl: I think it looks fine now 👍

schnittchen:
Nothing changed, only squashed

rhagenson (approved):
Looks good to me!

schnittchen:
Am I missing anything else?

ergl (approved):

ergl:
I don't think so, @rhagenson @EpicEric should this be merged?

rhagenson:
I do not see anything missing so it can be merged. Thank you @schnittchen for your work on this!

EpicEric (approved):

EpicEric:
Thank you @schnittchen!

PR #468, opened 2021-04-15
Update FFI chapters to reflect RFC 68

ergl:
Changes related to [RFC 68](https://github.com/ponylang/rfcs/blob/main/text/0068-force-declaration-of-ffi-functions.md)

### DIFF
--- a/docs/appendices/examples.md
+++ b/docs/appendices/examples.md
@@ -56,22 +56,26 @@ EGLEvent getEvent() {
 then you can destructure it and get the values using a tuple
 
 ```pony
+use @getEvent[EGLEvent]()
 type EGLEvent is (U8, F32, F32)
-(var code, var x, var y) = @getEvent[EGLEvent]()
+(var code, var x, var y) = @getEvent()
 ```
 
 ## Get and pass pointers to FFI
 
 ```pony
+use @XOpenDisplay[Pointer[_XDisplayHandle]](name: Pointer[U8] tag)
+use @eglGetDisplay[Pointer[_EGLDisplayHandle]](disp: Pointer[_XDisplayHandle])
+
 primitive _XDisplayHandle
 primitive _EGLDisplayHandle
 
-let x_dpy = @XOpenDisplay[Pointer[_XDisplayHandle]](U32(0))
+let x_dpy = @XOpenDisplay(Pointer[U8])
 if x_dpy.is_null() then
   env.out.print("XOpenDisplay failed")
 end
 
-let e_dpy = @eglGetDisplay[Pointer[_EGLDisplayHandle]](x_dpy)
+let e_dpy = @eglGetDisplay(x_dpy)
 if e_dpy.is_null() then
   env.out.print("eglGetDisplay failed")
 end
@@ -80,6 +84,9 @@ end
 ## Pass an Array of values to FFI (TODO)
 
 ```pony
+use @eglChooseConfig[U32](disp: Pointer[_EGLDisplayHandle], attrs: Pointer[U16] tag,
+  config: Pointer[_EGLConfigHandle], config_size: U32, num_config: Pointer[U32])
+
 primitive _EGLConfigHandle
 let a = Array[U16](8)
 a.push(0x3040)
@@ -92,7 +99,7 @@ a.push(0x3023)
 a.push(0x8)
 a.push(0x3024)
 let config = Pointer[_EGLConfigHandle]
-if @eglChooseConfig[U32](e_dpy, a, config, U32(1), Pointer[U32]) == 0 then
+if @eglChooseConfig(e_dpy, a.cpointer(), config, U32(1), Pointer[U32]) == 0 then
     env.out.print("eglChooseConfig failed")
 end
 ```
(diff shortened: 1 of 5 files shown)


ergl:
This is ready for review

docs/c-ffi/calling-c.md
```diff
@@ -97,20 +89,29 @@ let size = Winsize
...
-@os_send[U64](_event, data.cstring(), data.size()) ? // May raise an error
+use @printf[I32](fmt: Pointer[U8] tag, ...)
+// ...
+let run_ns: I64 = _current_t - _last_t
+let rate: I64 = (_partial_count.i64() * 1_000_000_000) / run_ns
+@printf("Elapsed: %lld,%lld\n".cstring(), run_ns, rate)
 ```
 
-If a signature declaration is used then that must be marked as possibly raising an error in the same way. The FFI call site must mark it as well.
+In the example above, the compiler will type-check the first argument to `printf`, but will not be able to check any other argument, since it lacks the necessary type information. It is __very__ important that you use `...` in the FFI signature if the corresponding C function is variadic: if you don't, the compiler might generate a program that is incorrect or crash, depending on the target platform.
```
jemc, line 104: Small suggestion to make it a bit more clear how creepy and subtle this problem can be:

```suggestion
In the example above, the compiler will type-check the first argument to `printf`, but will not be able to check any other argument, since it lacks the necessary type information. It is __very__ important that you use `...` in the FFI signature if the corresponding C function is variadic: if you don't, the compiler might generate a program that is incorrect or crash on some platforms while appearing to work correctly on others.
```

jemc (approved):

jemc (approved):

ergl:
@jemc I'm going to add some clarifying comments about what "compatible" type declarations really mean.

SeanTAllen:
@ergl does that mean i shouldnt merge this yet?

ergl:
@SeanTAllen No, not yet, I want to give others a chance to review the latest additions.

SeanTAllen:
@ergl who else do you want to review?

ergl:
@SeanTAllen if either you or @jemc or you can give a look at the latest section I've added, that would be great.

docs/appendices/examples.md
```diff
@@ -56,30 +56,17 @@ EGLEvent getEvent() {
...
-
-let x_dpy = @XOpenDisplay[Pointer[_XDisplayHandle]](U32(0))
-if x_dpy.is_null() then
-  env.out.print("XOpenDisplay failed")
-end
-
-let e_dpy = @eglGetDisplay[Pointer[_EGLDisplayHandle]](x_dpy)
-if e_dpy.is_null() then
-  env.out.print("eglGetDisplay failed")
-end
```
SeanTAllen, line 77: is this an intentional removal @ergl ?

docs/c-ffi/calling-c.md
```diff
@@ -2,44 +2,33 @@
...
-```pony
-@fwrite[U64](data.cstring(), U64(1), data.size(), _handle)
-```
-
-The main difference is the @ symbol before the function name. This is what tells us it's an FFI call. Any time you see an @ in Pony there's an FFI going on.
-
-The other key difference is that the return type of the function is specified after the function name, in square brackets. This is because the compiler needs to know what type the value returned is (if any), but has no way to determine that, so it needs you to explicitly tell it.
-
-There are a few unusual things going on with the arguments to this FFI call as well. For the second argument, for which we're passing the value 1, we've had to specify that this is a U64. Again this is because the compiler needs to know what size argument to use, but has no way to determine this.
-
```
SeanTAllen, line 16: whats the reason for the removal of this? At least out of context, this seems like important information.

docs/appendices/examples.md
ergl: Yes, this specific part is already covered here https://tutorial.ponylang.io/c-ffi/calling-c.html#get-and-pass-pointers-to-ffi, so it feels like a duplicate

docs/c-ffi/calling-c.md
ergl: My rationale is that by removing the syntax for writing out the return type of a FFI function at the call site and requiring a declaration, it's hard to show a one-line example, so I decided to skip the intro and start the chapter by showing how to write a declaration.

The paragraph about having to type `U64(1)` is not needed anymore, since the compiler can now infer the correct type by looking at the declaration. This means one is allowed to write `1` as the second argument without any problems.

The bit about the @ symbol is covered a bit later.

docs/appendices/examples.md
SeanTAllen: Gotcha. Looks reasonable.

docs/c-ffi/calling-c.md
SeanTAllen: Ok, I am ok with this. We can see if people have any issues with it.

SeanTAllen (approved):

SeanTAllen:
@ergl do you want @jemc to give another review as well?

SeanTAllen:
I added a DO NOT MERGE label as this shouldn't be merged until the corresponding changes to ponyc are released. I'll be doing the merging as part of the release process for the next ponyc.

ergl:
@SeanTAllen If he has the time to review the last commit, yes, but it's not a problem otherwise.

jemc (approved):

Issue #362, opened 2019-02-23
Object capabilities section of tutorial is anemic


SeanTAllen:
Has 

- no examples of code that use capabilities (file example, tcp example would both be good)
- has no examples of how those capabilities work under the hood
- doesn't really explain what "ambient authority" is
- doesn't show how to create your own capabilities for a library

EpicEric:
[Here's some helpful discussion](https://ponylang.zulipchat.com/#narrow/stream/189985-beginner-help/topic/AmbientAuth.20(and.20other.20auths.3F)) from Zulip that can be incorporated into this issue.

rhagenson:
Closed by #472

PR #472, opened 2021-04-26
Explain object capabilities in more detail

schnittchen:
Currently a very rough sketch for https://github.com/ponylang/pony-tutorial/issues/362

I hope I'm roughly on the right track here.


schnittchen:
@rhagenson can you help me with the failed build please?

I'd like to have a preview to be able to "zoom out" and see how I am doing...

rhagenson:
@schnittchen 

Getting it working on Netlify might be a bit difficult as it is failing with a 404 code on the edit link for the page you are adding. I want to look into if we can change the Netlify settings to ignore "internal" 404 codes in PRs as this will continue to be a problem anytime a page is added or renamed.

Do you have Docker available locally? I used a live preview of my work while initially working in the MkDocs Material conversion.

schnittchen:
Yes, I do... I didn't see any Dockerfile or docker-compose.yml, otherwise I would have tried that already... Or did I miss setup instructions?

rhagenson:
No, I think I forgot to add the local preview instructions...sorry. Give me a minute to get to my dev machine (on phone now).

rhagenson:
I see why I did not add the local live preview. Our use of plugins means I would need to add a Dockerfile that installs the additional plugins.

schnittchen:
@rhagenson if it's just a short snippet you could add it to CONTRIBUTING.md. Or just add the Dockerfile...

The only downside I see with adding the Dockerfile is that as long as it is not used in the CI build, there's a chance it might get out of sync over time

rhagenson:
I would prefer to get the PR preview working if possible as we use Material Insiders so a local build would have to use the free Material theme so would not show the true preview.

rhagenson:
The Dockerfile I have locally is:

```docker
FROM squidfunk/mkdocs-material
COPY requirements.txt .
RUN pip install --requirement requirements.txt
```

schnittchen:
I'll try that. If it works, I'll open a PR with that snippet too

SeanTAllen:
@rhagenson yeah those 404s failing during PRs are... well... we don't want to turn off entirely. but if they mean we can't get a preview.... better to turn them off.

rhagenson:
I then run the following to use it:

```bash
docker run -u 1000:1000 -p 8000:8000 --rm -it -v $(pwd):/docs <name of image>
```

schnittchen:
@rhagenson 

```
Step 3/3 : RUN pip install --requirement requirements.txt
 ---> Running in d56fed021b95
Collecting git+https://****@github.com/squidfunk/mkdocs-material-insiders.git (from -r requirements.txt (line 1))
  Cloning https://****@github.com/squidfunk/mkdocs-material-insiders.git to /tmp/pip-req-build-ik95_l99
  Running command git clone -q 'https://****@github.com/squidfunk/mkdocs-material-insiders.git' /tmp/pip-req-build-ik95_l99
  fatal: could not read Password for 'https://${INSIDERS_PAT}@github.com': No such device or address
WARNING: Discarding git+https://****@github.com/squidfunk/mkdocs-material-insiders.git. Command errored out with exit status 128: git clone -q 'https://****@github.com/squidfunk/mkdocs-material-insiders.git' /tmp/pip-req-build-ik95_l99 Check the logs for full command output.
ERROR: Command errored out with exit status 128: git clone -q 'https://****@github.com/squidfunk/mkdocs-material-insiders.git' /tmp/pip-req-build-ik95_l99 Check the logs for full command output.
WARNING: You are using pip version 21.0.1; however, version 21.1.1 is available.
You should consider upgrading via the '/usr/local/bin/python -m pip install --upgrade pip' command.
The command '/bin/sh -c pip install --requirement requirements.txt' returned a non-zero code: 1
```

rhagenson:
@SeanTAllen It is the same sort of issue we had for the 429 errors.

SeanTAllen:
@rhagenson it would be nice if github actions could have "partial failures". annotations are sort of for that, but i seriously doubt we can use from netlify or mkdocs. o well.

rhagenson:
Well then. If Sean is okay with it, I will add the necessary htmlproofer exclusion for 404s for GitHub addresses and then the preview should ignore those internal 404s.

SeanTAllen:
@rhagenson i think the most important thing is the deploy-previews. the 404 checking is a nice to have. whatever you need to do, we accept the tradeoff and maybe one day we can do better.

schnittchen:
Maybe move the 404 checking to an extra step?

rhagenson:
I am leaning toward ignore the 404 errors on github. I wonder if I can ignore only 404s on our edit pages? I will reread the htmlproofer implementation.

Moving 404 checking to a separate step is not really an option as it fails on htmlproofer (as it should), but the extra step would have to occur in htmlproofer itself.

rhagenson:
I am working on the changes now. I realized that with the way htmlproofer checks for exclusions we get every combination (i.e., ignoring 404s on all urls listed).

rhagenson:
That is the best I think we can do right now. Rather than ignoring 404s and 429s across all of GitHub it should only apply to ponylang repositories.

schnittchen:
Thank you @rhagenson! I'll rebase later to pick up your config change.

rhagenson:
Sounds good. Good luck with writing! Let me know if I can help further.

I have been asked to open an issue on htmlproofer that will reduce 429s so I wonder if they would be willing to accept a similarly important "don't exclude all combinations" issue? Only one way to find out!

schnittchen:
There's still stuff lying around: the big block comment at the top is my initial checklist and will be deleted later.

The TOC entry does not yet coincide with the h1.

Other than that, it's review time!

schnittchen:
It's been a while. While there's more work to do, can I get a first review to see if I got the general stuff right?

docs/object-capabilities/derived-authority.md
```diff
@@ -0,0 +1,109 @@
...
+to be authorized, and allows us to restrict and delegate this authority inside our program.
+
+Recall the definition of capability from [ref]:
+
+> A capability is an unforgeable token that (a) designates an object and (b) gives the program the authority to perform a specific set of actions on that object.
+
+In Pony, the Main actor is passed the environment object, which keeps an `AmbientAuth` in the `root` field. This value is the capability that represents the ambient authority given to us by the system.
+Instead of allowing unfettered access to the outside world, Pony enforces the use of this authority.
+
+Here is a simple program that connects to example.com via TCP on port 80 and quits:
```
rhagenson, line 27: Drop "simple" please. It is a rather simple program, but as someone learns, seeing a word like "simple" then not fully understanding a program like this one can be discouraging.

docs/object-capabilities/derived-authority.md
```diff
@@ -0,0 +1,109 @@
...
+actor Connect
+  new create(out: OutStream, auth: AmbientAuth) =>
+    TCPConnection(auth, MyTCPConnectionNotify(out), "example.com", "80")
+
+actor Main
+  new create(env: Env) =>
+    try Connect(env.out, env.root as AmbientAuth) end
+```
+
+The Main actor authorizes the Connect actor by passing the ambient authority token on to it. This token is not forgeable since the `AmbientAuth` constructor is private—only the runtime glue code
```
rhagenson, line 54: Format `Main` and `Connect` please.

The only previous mention of token in this section is in the quote from the previous section. Can you work a mention of "ambient authority token" into this section earlier prior to the example?

Please use "unforgeable" rather than "not forgeable" so we state a consistent negation on forgeability.

Stating the runtime as "glue code" sounds dismissive to me. I understand what you mean though. Can you change it to something like: "This `AmbientAuthority` token is unforgeable since it is provided by the runtime itself so is guaranteed to have only one instance."

docs/object-capabilities/derived-authority.md
```diff
@@ -0,0 +1,109 @@
...
+can create the one and only instance.
+
+The Connect actor uses this authority when it creates a TCP connection:
```
rhagenson, line 57: `Connect` monospace/code formatting please.

docs/object-capabilities/derived-authority.md
```diff
@@ -0,0 +1,109 @@
...
+
+```pony
+TCPConnection(auth, MyTCPConnectionNotify(out), "example.com", "80")
+```
+
+The TCPConnection requires an authority as first parameter, and since the compiler checks that
```
rhagenson, line 63: `TCPConnection` monospace/code formatting please.

docs/object-capabilities/derived-authority.md
```diff
@@ -0,0 +1,109 @@
...
+the correct type was passed, this guarantees that a TCPConnection can only be created by an
```
rhagenson, line 64: `TCPConnection` monospace/code formatting please.

docs/object-capabilities/derived-authority.md
```diff
@@ -0,0 +1,109 @@
...
+actor holding the required authorization. If you look into the implementation of the `TCPConnection`
+constructor, you will notice that the authorization is not even used beyond the declaration of the
+parameter.
```
rhagenson, line 67: I am hesitant to suggest looking at the implementation of `TCPConnection` at this stage and especially hesitate to state "you will notice that the authorization is not even used beyond the declaration of the parameter." unless we explain **why** that is in the same location. Such an instance of taking a parameter only to not even use it will be, at the very least, odd to many programmers -- beginners and experiences alike.

docs/object-capabilities/derived-authority.md
```diff
@@ -0,0 +1,109 @@
...
+
+## Restrict, then delegate your authority
+
+In order to handle our own code and that of others more safely, and also to understand our code better,
+we want to split up the authority, and only grant the particular authority a piece of code actually
+requires.
```
rhagenson, line 73: I think you know this, but as written this is not quite why we restrict authority. There is an added benefit of code readability and understandability, but the real reason we restrict and delegate is to minimize capabilities to only those which are necessary to safely accomplish the task at hand. Passing/allowing more capabilities than are necessary is possibly unsafe.

docs/object-capabilities/derived-authority.md
```diff
@@ -0,0 +1,109 @@
...
+
+Imagine we don't trust the `Connect` actor, so we don't want to provide it with more authority
```
rhagenson, line 75: Would you mind changing the "don't"'s to "do not"'s? Doing so should make it slightly easier to read for those without English as a primary language at the cost of reading awkwardly to primary English speakers.

docs/object-capabilities/derived-authority.md
```diff
@@ -0,0 +1,109 @@
...
+than needed. For example, there is no point in granting it filesystem access.
```
rhagenson, line 76: You have an implicit assumption here that the reader knows the `TCPConnection` does not need filesystem access. Please clarify what access/capabilities `TCPConnection` needs and does not need closer to the example above.

docs/object-capabilities/derived-authority.md
```diff
@@ -0,0 +1,109 @@
...
+
+The first parameter of the `TCPConnection` constructor has the type
+
+```pony
+type TCPConnectionAuth is (AmbientAuth | NetAuth | TCPAuth | TCPConnectAuth)
+```
+
+This looks like a nice hierarchy of authority, in order of decreasing rights. We can pass
```
rhagenson, line 84: I do not think the hierarchy of authority is obvious here. Rather than stating it looks like a nice hierarchy, can you change this to explain that it is a nice hierarchy of decreasing authority?

docs/object-capabilities/derived-authority.md
```diff
@@ -0,0 +1,109 @@
...
+  new create(out: OutStream, auth: TCPConnectionAuth) =>
+    TCPConnection(auth, MyTCPConnectionNotify(out), "example.com", "80")
+
+actor Main
+  new create(env: Env) =>
+    try Connect(env.out, TCPConnectAuth(env.root as AmbientAuth)) end
+```
+
+Now we are sure it cannot access the filesystem or listen on a TCP or UDP port. This is possible
+because there is a `TCPConnectAuth` constructor which accepts the AmbientAuth.
```
rhagenson, line 98: Will you textually highlight the changes made here? Then explain that by switching from accepting `AmbientAuth` to `TCPConnectAuth` we lose the ability to access the filesystem -- exactly as we want set out to do with the changes made here.

docs/object-capabilities/derived-authority.md
```diff
@@ -0,0 +1,109 @@
...
+
+You can see this pattern over and over in the standard library.
```
rhagenson, line 100: Remove this please. The subsection does not need to end with a statement about the standard library as it is not about the standard library.

docs/object-capabilities/derived-authority.md
```diff
@@ -0,0 +1,109 @@
...
+
+## Authorization-friendly interface
+
+Consider the above example again, but this time let's treat the `Connect` actor as coming from
+a 3rd party package. If the author of that package had not considered which authority exactly
+was necessary, it would not be possible to restrict the authoritization given to its code.
+A function with an `AmbientAuth` parameter will not accept a lesser auth token.
+
+Therefore it is necessary to design your packages and functions with this in mind.
```
rhagenson, line 109: Can you logically flip this around? By that I mean rather than placing the reader as using an outside package, place them as writing the outside package. This edit is a preference request as there is nothing wrong with how it is written, but I think switching it around will let you more cleanly and clearly jump off from the previous subsection where we made a change to the **implementation** rather than a change to how the implementation is used.

rhagenson (changes requested):
Overall, I think you did get the general stuff right. I made a number of suggestions for motivating why we manage capabilities rather than have ambient authority everywhere.

Did you have any specific concerns that I did not address in my suggestions?

docs/object-capabilities/derived-authority.md
```diff
@@ -0,0 +1,109 @@
...
+> * [ ] come up with a better title for the section
+> * [ ] toc stuff and linking
+
+Any interesting program will need to interact with the outside world by using operating system concepts, like accessing the network or the file system, or by creating and communicating with other programs.
+The operating system has granted the right to do so to the program at startup (but it might do some checks at runtime, for example controlling access to a file). This is called __ambient authority__.
+
+The operating system has no concept of how our program is composed inside, and it knows nothing about Pony specific concepts in particular. Pony requires operations which interact with the system
+to be authorized, and allows us to restrict and delegate this authority inside our program.
+
+Recall the definition of capability from [ref]:
```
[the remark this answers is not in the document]
schnittchen: Yes, that's the idea. I'm not 100% sure about the title of this section and the ordering though, so I kept it open

docs/object-capabilities/derived-authority.md
```diff
@@ -0,0 +1,109 @@
...
+
+> A capability is an unforgeable token that (a) designates an object and (b) gives the program the authority to perform a specific set of actions on that object.
+
+In Pony, the Main actor is passed the environment object, which keeps an `AmbientAuth` in the `root` field. This value is the capability that represents the ambient authority given to us by the system.
```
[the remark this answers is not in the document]
schnittchen: It does not mention the Pony runtime which sets up the env and instantiates the actor, but mentioning that here would be too specific I believe. What's the best way to refer to the `env: Env`passed to `Main` without being too technical but also not too vague?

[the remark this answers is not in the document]
rhagenson: Good question. Not sure I have a good answer. Perhaps "In Pony, the Main actor is created with an `Env` object, which holds the unforgeable `AmbientAuth` token in its `root` field." Rather using "pass" which suggests some transfer, use the terminology of Pony itself where the `Env` is required in something call `create` -- best solution I can think of which avoid being vague while also not having to get into the runtime's role in it all.

docs/object-capabilities/derived-authority.md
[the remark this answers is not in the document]
rhagenson: Sounds good. I wanted to make sure there was no plan to reference that from elsewhere.

docs/object-capabilities/derived-authority.md
schnittchen: I think I'll move that aspect into a separate section. It's interrupting the flow of understanding here.

schnittchen:
I apologize for the loooong delay. I ran out of energy for some time...

docs/object-capabilities/derived-authority.md
```diff
@@ -0,0 +1,109 @@
...
+> * show how a derived authority can be constructed
+> * refer to the patterns for more complex scenarios
+>
+> ## [TODO]
+>
+> * [ ] come up with a better title for the section
+> * [ ] toc stuff and linking
+
+Any interesting program will need to interact with the outside world by using operating system concepts, like accessing the network or the file system, or by creating and communicating with other programs.
+The operating system has granted the right to do so to the program at startup (but it might do some checks at runtime, for example controlling access to a file). This is called __ambient authority__.
```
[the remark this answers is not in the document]
schnittchen: I'm not sure I actually did what you asked me for here and in the previous comment, but it feels better now. Please have another look

docs/object-capabilities/derived-authority.md
schnittchen: I'm not sure I really made it clearer here

docs/object-capabilities/derived-authority.md
schnittchen: How would I do that, the highlighting?
I made textual changes that should make it clearer.

docs/object-capabilities/derived-authority.md
schnittchen: I did that and also expanded a bit. Not sure the introduction of another level of type aliasing is really useful though.

docs/object-capabilities/derived-authority.md
```diff
@@ -0,0 +1,147 @@
...
+We call __ambient authority__ all those rights implicitly granted to the program to make these things possible, and because Pony makes this concept *explicit*, we need to take
+the time to talk about what it means and how it works. In other languages like for example C, you can always attempt to do an operation like I/O, and it will usually succeed
+save some runtime checks (your disk being full might make an operation fail, for example). But in Pony, any piece of code interacting with the outside world needs the authority to do so.
+
+The operating system essentially knows nothing about the structure of any program it runs, in particular it is ignorant of any Pony specific concepts.
+In order to impose that some code is authorized for an operation like writing to disk, Pony requires a special argument to be passed, a capability bearing the authority required for the
+task at hand. The type system guarantees that no code that has not been passed this capability can perform the task requiring authorization, so this kind of safety is guaranteed at compile time.
```
rhagenson, line 21:
```suggestion
task at hand. The type system guarantees that inadequate capabilities for a task fail at compile time.
```

Purpose here being to remove the double-negative of "**no** code that has **not** been passed".

docs/object-capabilities/derived-authority.md
rhagenson:
```suggestion
The `Connect` actor uses this authority when it creates a TCP connection:
```

docs/object-capabilities/derived-authority.md
rhagenson:
```suggestion
The `TCPConnection` requires an authority as first parameter, and since the compiler checks that
```

docs/object-capabilities/derived-authority.md
rhagenson:
```suggestion
the correct type was passed, this guarantees that a `TCPConnection` can only be created by an
```

docs/object-capabilities/derived-authority.md
```diff
@@ -0,0 +1,147 @@
...
+
+```pony
+type TCPConnectionAuth is (AmbientAuth | NetAuth | TCPAuth | TCPConnectAuth)
+```
+
+you might notice that this looks like a hierarchy of authorities:
+
+`AmbientAuth >> NetAuth >> TCPAuth >> TCPConnectAuth`
+
+where in this paragraph, ">>" means "grants at least as much authority as". In fact, the AmbientAuth
```
rhagenson, line 137:
```suggestion
where in this paragraph, ">>" means "grants at least as much authority as". In fact, the `AmbientAuth`
```

rhagenson (commented):
@schnittchen Thank you for getting back to this. I have offered some comments and suggestions for improvement.

I am also going to review my previous comments and resolve any I think have been sufficiently addressed or otherwise are not longer relevant.

docs/object-capabilities/derived-authority.md
rhagenson:
```suggestion
The `Main` actor authorizes the `Connect` actor by passing the ambient authority token on to it. This token is not forgeable since the `AmbientAuth` constructor is private—only the runtime glue code
```

docs/object-capabilities/derived-authority.md
schnittchen: In addition it talks about passing the wrong capabiity, where I only thought about "none". Thanks!

docs/object-capabilities/derived-authority.md
```diff
@@ -0,0 +1,147 @@
...
+>
+> ## [TODO]
+>
+> * [ ] come up with a better title for the section
+> * [ ] toc stuff and linking
+
+Any interesting program will need to interact with the outside world, like accessing the network or the file system, or by creating and communicating with other programs.
+We call __ambient authority__ all those rights implicitly granted to the program to make these things possible, and because Pony makes this concept *explicit*, we need to take
+the time to talk about what it means and how it works. In other languages like for example C, you can always attempt to do an operation like I/O, and it will usually succeed
+save some runtime checks (your disk being full might make an operation fail, for example). But in Pony, any piece of code interacting with the outside world needs the authority to do so.
```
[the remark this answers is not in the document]
schnittchen: Thanks!

docs/object-capabilities/derived-authority.md
schnittchen: I think this has been solved

schnittchen:
I think this could be close to be done up to "technical" stuff like adding it properly to the TOC.
I've added a line to `mkdocs.yml`, but that does not seem to work (the netlify build step fails with a 404?). I've removed it but then I get an additional warning...

I still need to decide on a title, for that it would be nice to read it in a browser and see it in context.

Also, the spell checker seems not to know a few existing words (like filesystem), is there a config for adding those?

SeanTAllen:
@schnittchen you can add additional words to the dictionary by modifying `.spelling-wordlist.txt`

when adding anything, please make sure it is in alphabetical order.

rhagenson:
I looked at why the PR preview is failing and it seems to be caused by a mismatch between how we expect the htmlproofer exclusions to work and how they do work. I mean for the exclusion to partially match URL, but instead it must match fully.

This is causing our edit pages to fail on new additions because the page does not yet exist on main and therefore there is no edit page yet.

My proposed solutions are:

1. Submit another feature request to mkdocs-htmlproofer to have partial URL match
2. Remove mkdocs-htmlproofer
3. Replace mkdocs-htmlproofer

Htmlproofer is serving its purpose despite it not working as we intend. It is supposed to tell us when a link is dead or otherwise problematic. It just is a problem now because we are in effect creating a link which will not be dead after the PR is merged.

schnittchen:
Hmm, none of these steps is directly connected to this PR. Is there a way to move this forward?

rhagenson:
@schnittchen Getting the preview working with adding pages insan issue I will need to address. As for this PR, what I see at this point are a few formatting changes (e.g., Main to `Main` and Connect to `Connect`) and removing your Plan/TODO content before it can be merged.

rhagenson:
@schnittchen Is there anything else you needed from me to move this PR forward?

docs/object-capabilities/derived-authority.md
schnittchen: I think I solved most if not all of this now.

Regarding the use of "token", it's mentioned again right below the quote, in `which holds the unforgeable `AmbientAuth` token in its root field`. Is that enough to guide the reader?

schnittchen:
> @schnittchen Is there anything else you needed from me to move this PR forward?

Thanks, I followed up on the remaining points, I hope this is now really close to done.

I removed my TODO at the top, and before merging I'd squash everything into a single commit.

docs/object-capabilities/derived-authority.md
```diff
@@ -0,0 +1,138 @@
...
+actor Connect
+  new create(out: OutStream, auth: TCPConnectionAuth) =>
+    TCPConnection(auth, MyTCPConnectionNotify(out), "example.com", "80")
+
+actor Main
+  new create(env: Env) =>
+    try Connect(env.out, TCPConnectAuth(env.root as AmbientAuth)) end
+```
+
+Now we are sure it cannot access the filesystem or listen on a TCP or UDP port.
```
rhagenson, line 93: This sentence confused me on re-reading as I often forget the hierarchy of authorities in net. I do not think it should block merging this, but perhaps before the code block when talking about "downgrading" a brief little rewrite might help?

> ...`AmbientAuth`, we can "downgrade" that to a `TCPConnectAuth`,...

to

> ...`AmbientAuth` (the root of all authority), we can "downgrade" that to a `TCPConnectAuth` (the most restrictive authority in net),...

docs/object-capabilities/derived-authority.md
```diff
@@ -0,0 +1,138 @@
...
+## Authorization-friendly interface
+
+Consider the above example again, but this time let's think of the `Connect` actor being part
+of a 3rd party package that we are building. Our goal is to write the actor in such a way that
+users of our package can grant it only the authority necessary for it to function.
+
+As the package author, it is then our responsibility to realize that the minimal authority
+possible is the `TCPConnectAuth`. But anything that the `TCPConnection` constructor accepts
+is valid, and a user of our package might be strict and pass a `TCPConnectionAuth`, or might
+be lax and just pass the `AmbientAuth`.
```
rhagenson, line 104: `TCPConnectionAuth` includes `AmbientAuth` -- did you mean `TCPConnectAuth`?

docs/object-capabilities/derived-authority.md
rhagenson: You explain this further below so not much is needed, but contextualizing the "downgrade" might make it easier to understand on first read.

rhagenson (commented):
I left a few minor comments, but otherwise I think this is ready to merge. Only possibly necessary change is an instance of `TCPConnectionAuth` when I think `TCPConnectAuth` was meant.

Thank you for continuing to push this forward @schnittchen! This is a fantastic and much needed addition to the Tutorial!

docs/object-capabilities/derived-authority.md
schnittchen: Sure thing!

PR #473, opened 2021-04-30
Document consume with capability

### LINKED ISSUE #470: consume with capability should be documented
Currently we cover `consume` on the [Consume and Destructive Read](https://tutorial.ponylang.io/reference-capabilities/consume-and-destructive-read.html) section, but the text only uses it in the context of assigning the result to an `iso` variable:

```pony
fun test(a: Wombat iso) =>
  var b: Wombat iso = consume a // Allowed!
```

We should mention that `consume` accepts an optional capability, like `consume val` or `consume ref`, although you can omit the capability and annotate the variable you're assigning:

```pony
let b: Wombat val = consume b 
let b = consume val b // Same as above
```

This is similar to what can be done with `recover`, so I'm unsure if this should go in the "Consume and Destructive Read" section or in the "Recovering Capabilities" section.

ergl:
Closes #470

### DIFF
--- a/docs/reference-capabilities/consume-and-destructive-read.md
+++ b/docs/reference-capabilities/consume-and-destructive-read.md
@@ -23,6 +23,15 @@ fun test(a: Wombat iso) =>
 
 Here's an example of that. When you try to assign `a` to `c`, the compiler will complain.
 
+By default, a `consume` expression returns a type with the capability of the variable that you are assigning to. You can see this in the example above, where we say that `b` is `Wombat iso`, and as such the result of the `consume` expression is `Wombat iso`. We could also have said that `b` is a `Wombat val`, but sometimes, like for example if the type of a variable is quite long, we can instead give an explicit reference capability to the `consume` expression:
+
+```pony
+fun test(a: AnIncrediblyLongTypeName iso) =>
+  var b = consume val a
+```
+
+The example above is equivalent to saying `var b: AnIncrediblyLongTypeName val = consume a`.
+
 __Can I `consume` a field?__ Definitely not! Consuming something means it is empty, that is, it has no value. There's no way to be sure no other alias to the object will access that field. If we tried to access a field that was empty, we would crash. But there's a way to do what you want to do: _destructive read_.
 
 ## Destructive read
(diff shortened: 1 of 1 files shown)


docs/reference-capabilities/consume-and-destructive-read.md
```diff
@@ -23,6 +23,15 @@ fun test(a: Wombat iso) =>
 
 Here's an example of that. When you try to assign `a` to `c`, the compiler will complain.
 
+By default, a `consume` expression returns a type with the capability of the variable that you are assigning to. You can see this in the example above, where we say that `b` is `Wombat iso`, and as such the result of the `consume` expression is `Wombat iso`. We could also have said that `b` is a `Wombat val`, but sometimes, like for example if the type of a variable is quite long, we can instead give an explicit reference capability to the `consume` expression:
```
EpicEric, line 26: Not sure if that particular motive should be expressed. There could be other reasons to opt for this sugar.

```suggestion
By default, a `consume` expression returns a type with the capability of the variable that you are assigning to. You can see this in the example above, where we say that `b` is `Wombat iso`, and as such the result of the `consume` expression is `Wombat iso`. We could also have said that `b` is a `Wombat val`, but we can instead give an explicit reference capability to the `consume` expression:
```

However, I don't think the example needs to be changed.

docs/reference-capabilities/consume-and-destructive-read.md
```diff
@@ -23,6 +23,15 @@ fun test(a: Wombat iso) =>
...
+
+```pony
+fun test(a: AnIncrediblyLongTypeName iso) =>
+  var b = consume val a
+```
+
+The example above is equivalent to saying `var b: AnIncrediblyLongTypeName val = consume a`.
```
EpicEric, line 33:
```suggestion
The expression in line 2 of the example above is equivalent to saying `var b: AnIncrediblyLongTypeName val = consume a`.
```

EpicEric (changes requested):

docs/reference-capabilities/consume-and-destructive-read.md
ergl: Sounds good, I've removed it

EpicEric (approved):

rhagenson (approved):
Looks good to me!

Issue #466, opened 2021-03-29
FFI: Using Pointer[None] for void* parameters should be documented


ergl:
When trying to come up with a valid FFI declaration for `memcpy`, I struggled finding a good type for the parameters. Take the following example:

```pony
let out: Pointer[Pointer[U8] tag] tag = // ...
let outlen: Pointer[U8] tag = // ...
let ptr: Pointer[U8] tag = // ...
let size: USize = // ...
// ...
@memcpy[None](out, addressof ptr, size.bitwidth() / 8)
@memcpy[None](outlen, addressof size, USize(1))
```

The first usage of `memcpy` has types `Pointer[Pointer[U8] tag] tag` for the first two parameters, while the second one has `Pointer[U8] tag` for the first one, and `Pointer[USize]` for the second one.

On first look, it seems hard to write a single declaration that fits both usages. But the compiler allows using the `Pointer[None]` type as a ["catch-all" pointer type](https://github.com/ponylang/ponyc/blob/6d1080eb6237f76c74b021f324e5f20d13accae5/src/libponyc/expr/ffi.c#L10-L34), such that any FFI call with `Pointer[_]`, `NullablePointer[_]` or `USize` argument will be accepted as a valid type. As such, the above code is valid with the following FFI declaration:

```pony
use @memcpy[Pointer[U8]](dst: Pointer[None], src: Pointer[None], n: USize)
```

redvers:
Could @free[]() have similar concerns?

I just checked - the special treatment doesn't seem to apply to structs.  Should it?

```
actor Main
  new create(env: Env) =>
    env.out.print("Foo")

    var wibble: Foo = Foo
    wibble.bar = "Hello".cstring()
//  wibble.baz = "Hello".cstring()

//  var wibble2: Pointer[U8] tag = wibble.baz
    var wibble3: Pointer[None] tag = wibble.baz


struct Foo
  var foo: U64 = U64(0)
  var bar: Pointer[U8] tag = Pointer[U8]
  var baz: Pointer[None] tag = Pointer[None]
```

redvers:
(I'm guessing not, but mentioning it just in case)

ergl:
@redvers Yes, using this for `@free` would work. As for structs, it seems like this check only applies when type-checking FFI calls against FFI declarations. The following example compiles and would allow you to pass any of the `Foo` fields to free.

```pony
use @free[None](arg: Pointer[None])

actor Main
  new create(env: Env) =>
[...]

Issue #332, opened 2018-11-24
The FFI-chapter needs some clarification on addressof and C


marsuk:
In Chapter 9 "C FFI" [](https://tutorial.ponylang.io/c-ffi/calling-c.html) in the paragraph "C types", after saying "just like taking an address in C" it might be good to say something like this:

Note that in C, when you write only the variable name of a struct instance, C's address operator & gives you the pointer to the first member. This is not the case in Pony. You have to state the first member of the struct, otherwise it results in a pointer that points to the pointer to the first member.

jemc:
I don't really follow the particular example text you gave.

Can you elaborate on what you think is missing from the original description?

marsuk:
It's just that Pony doesn't do what you would expect when you have used C before.

In C: `&struct_instance` == `&struct_instance.first_member`
But in Pony: `adressof struct_instance` != `adressof struct_instance.first_member`

This means, in FFI Pony code you currently cannot just write `addressof window_class` or anything like that, because it won't work. I had to write `addressof window_class.cbSize` to get the correct address of the structure. `cbSize` is the first field in the structure.

The address that you get with `addressof window_class` is something unrelated, maybe something that Pony uses internally. It happens to be another pointer pointing to the what I actually wanted. Maybe mentioning this in the tutorial would be too confusing, or this ponyc behaviour might change anyway.

When the behaviour of Pony's `addressof` is what you want it to be and is not a bug, for whatever reason, then it would be good to say in the tutorial that Pony users need to be aware of it and that it will not look like C code looks like.

SeanTAllen:
Can you supply a small snippet code that can be used to see what you are talking about @marsuk ?

marsuk:
Yes, in the issue https://github.com/ponylang/ponyc/issues/2944 I was asking how I would eventually create a Windows application, and there is code like this. Although, that issue was not about this problem here.

That's the shorter one of the two versions in the issue:
https://gist.github.com/ponylang-gist/4ee25805123c8c646ce5fc6a1db93975

When you change line 97 from
`let class_atom: ATOM = @RegisterClassExA[ATOM](addressof wndclass.cbSize)`
to
`let class_atom: ATOM = @RegisterClassExA[ATOM](addressof wndclass) // .cbSize)`
the application will not start anymore.

ergl:
@SeanTAllen I think this can be closed since it was partially fixed by #474, which added the following:

```
[...]

Issue #303, opened 2018-06-20
Update FFI docs to include how to call "reserved words" functions


SeanTAllen:
See https://github.com/ponylang/ponyc/issues/2789 where you can call a c-function called box via @"box" whereas if you use @box you get a compilation error. This should be documented in the tutorial in the C-ffi section.

ergl:
Closed on #474

Issue #82, opened 2016-04-06
Improve FFI documentation


SeanTAllen:

aturley:
I added an issue (#83) for wanting a section about callbacks, but maybe that should be in here?

SeanTAllen:
I think that is reasonable for now

MrMormon:
[These](https://bluishcoder.co.nz/2015/11/04/a-quick-look-at-pony.html) [resources](https://bluishcoder.co.nz/2015/12/16/c-linkable-libraries-with-pony.html) talk about, among other things, how to statically and dynamically link C libraries. I'm not sure which parts are outdated/wrong now, if any. I'm uncertain enough as it is whether I can make [JUCE](https://www.juce.com/discover) work with Pony, since it's C++ and people are expected to know the GCC/Clang flags to use.

PR #474, opened 2021-05-11
Improve FFI documentation

### LINKED ISSUE #263: The FFI section should be revised to include more details about the C/Pony API and ABI compatibility and good practices
This is a list of things that either are imprecise or are bothering me in the current FFI section of the tutorial. Discussion is open on whether the good practices I'm going to recommend really are good practices.

1. Clarify the C/Pony type mapping. In particular, the tutorial should make it clear that Pony classes and structs are reference types equivalent to C pointer-to-structs and that only Pony structs can be used to represent non-opaque C structs.
2. State that passing things by value in the FFI isn't possible at the moment, but that it is something we intend to support in the future.
3. Recommend `Pointer[None]` for `void*` instead of `Pointer[U8]` since `Pointer[None]` basically has `void*` semantics in FFI parameters, where it is a supertype of every ABI-compatible type.
4. Detail why pointers to primitives are used for opaque types instead of `Pointer[None]`. This doesn't change anything at the ABI level but provides some amount of type safety. In addition, state that the associated primitives shouldn't be used anywhere except as type parameters to `Pointer`. This avoids any misuse of the FFI.
5. Clarify the usage of `MaybePointer` and what it means for the API and ABI. It is only useful as a return type or an output parameter type (possibly nested in a struct in both cases), where receiving a null value on the Pony side is a possibility. It is unnecessary as an input parameter type.

ergl:
This is a general stab at making the FFI docs better / clarifying some details. Hopefully the commit messages clarify individual changes.

Fixes #263, #303 and #466.
Resolves some of the points raised in #82, but maybe it should be kept open as a "catch-all" FFI docs improvement issue.

And I think #332 is also fixed by this, by clarifying that structs are equivalent to pointer to structs from the C point of view.

### DIFF
--- a/docs/c-ffi/calling-c.md
+++ b/docs/c-ffi/calling-c.md
@@ -42,7 +42,23 @@ In Pony, a String is an object with a header and fields, while in C a `char*` is
 
 Pony classes and structs correspond directly to pointers to the class or struct in C.
 
-For C pointers to simple types, such as U64, the Pony `Pointer[]` polymorphic type should be used, with a `tag` reference capability. `Pointer[U8] tag` should be used for void*.
+For C pointers to simple types, such as U64, the Pony `Pointer[]` polymorphic type should be used, with a `tag` reference capability. To represent `void*` arguments, you should the `Pointer[None] tag` type, which will allow you to pass a pointer to any type, including other pointers. This is needed to write declarations for certain POSIX functions, such as `memcpy`:
+
+```pony
+// The C type is void* memcpy(void *restrict dst, const void *restrict src, size_t n);
+use @memcpy[Pointer[U8]](dst: Pointer[None] tag, src: Pointer[None] tag, n: USize)
+
+// Now we can use memcpy with any Pointer type
+let out: Pointer[Pointer[U8] tag] tag = // ...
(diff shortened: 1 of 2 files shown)


docs/c-ffi/calling-c.md
```diff
@@ -71,11 +91,53 @@ if e_dpy.is_null() then
...
+
+**Can I pass struct types by value, instead of passing a pointer?** Not at the moment. This is a known limitation of the current FFI system, but it is something the Pony team is interested in fixing.
 
-A common pattern in C is to pass a struct pointer to a function, and that function will fill in various values in the struct. To do this in Pony, you make a `struct` and then use a `NullablePointer`:
+### Working with Structs: from C to Pony
+
+A common pattern in C is to pass a struct pointer to a function, and that function will fill in various values in the struct. To do this in Pony, you make a `struct` and then use a `NullablePointer`, which denotes a possibly-null type:
 
 ```pony
+use @ioctl[I32](fd: I32, req: U32, ...)
```
ergl, line 139: I'm not too happy with this example, since it's a variadic function, wich we only mention later on. I tried finding a good example of NullablePointer, but couldn't come up with a real, simple example.

PR #475, opened 2021-05-11
The mandatory FFI "use" change improves our safety story.

redvers:
Even though it is true that we can't make guarantees around FFI calls, the pony type system coupled with the now mandatory "use" statements does provide all the tools a programmer needs to guarantee that the data that they pass into their FFI functions is correct.

(Where "correct" is left as an exercise to the reader as they are the ones that need to strictly define what is "correct".  We can't automatically validate either the "shape" of the data or derive who manages the memory).

The point being, if they correctly describe the data using pony types, the pony compiler *will* enforce the shape of that data for them consistently throughout their application.  That is *huge*.

(How to map pony-types and build type-systems that are flexible _and_ safe is 95% of the difficulty of doing C-FFI)

### DIFF
--- a/docs/types/at-a-glance.md
+++ b/docs/types/at-a-glance.md
@@ -35,4 +35,4 @@ The Pony type system offers a lot of guarantees, even more than other statically
 
 Some of those will make sense right now. Some of them may not mean much to you yet (like capabilities-security and causal messaging), but we'll get to those concepts later on.
 
-__If I use Pony's FFI to call code written in another language, does Pony magically make the same guarantees for the code I call?__ Sadly, no. Pony's type system guarantees only apply to code written in Pony. Code written in other languages gets only the guarantees provided by that language.
+__If I use Pony's FFI to call code written in another language, does Pony magically make the same guarantees for the code I call?__ Sadly, no. Pony has no way of determining if the FFI definition you provide is in the format your FFI library expects. Pony's type system does guarantee that data passed to your FFI calls match those type definitions you provide. This additional safety is not a guarantee. Pony's type system can only guarantee code written in Pony. Code written in other languages gets only the guarantees provided by that language.
--- a/docs/types/at-a-glance.md
+++ b/docs/types/at-a-glance.md
@@ -39,4 +39,4 @@ __If I use Pony's FFI to call code written in another language, does Pony magica
 
 Pony requires FFI parameters to be explicitly defined as Pony types in advance. This allows the typechecker to guarantee that no code-paths exist that would result in an FFI call that violated the top-level definition. This assurance is not available in many other languages.
 
-Unfortunately, As there is no way Pony can validate compatibility between your top-level definition and what the library expects Pony's guarantees cannot be extended.
+Unfortunately, as there is no way Pony can validate compatibility between your top-level definition and what the library expects, Pony's guarantees cannot be extended.
--- a/docs/types/at-a-glance.md
+++ b/docs/types/at-a-glance.md
@@ -37,7 +37,6 @@ Some of those will make sense right now. Some of them may not mean much to you y
 
 __If I use Pony's FFI to call code written in another language, does Pony magically make the same guarantees for the code I call?__ Sadly, no. Pony's type system can only guarantee code written in Pony. Code written in other languages gets only the guarantees provided by that language.
 
-Pony requires FFI parameters be explicitly defined as Pony types in advance. This allows the typechecker to guarantee that no code-paths exist that would result in an FFI call that violated the top-level definition. This assurance is not available in many other languages.
+Pony requires FFI parameters to be explicitly defined as Pony types in advance. This allows the typechecker to guarantee that no code-paths exist that would result in an FFI call that violated the top-level definition. This assurance is not available in many other languages.
 
 Unfortunately, As there is no way Pony can validate compatibility between your top-level definition and what the library expects Pony's guarantees cannot be extended.
-
--- a/docs/types/at-a-glance.md
+++ b/docs/types/at-a-glance.md
@@ -35,8 +35,9 @@ The Pony type system offers a lot of guarantees, even more than other statically
 
 Some of those will make sense right now. Some of them may not mean much to you yet (like capabilities-security and causal messaging), but we'll get to those concepts later on.
 
-__If I use Pony's FFI to call code written in another language, does Pony magically make the same guarantees for the code I call?__ Sadly, no. Pony's type system can only guarantee code written in Pony. Code written in other languages gets only the guarantees provided by that language.
+__If I use Pony's FFI to call code written in another language, does Pony make the same guarantees for the code I call?__ Sadly, no. Pony's type system can only guarantee code written in Pony. Code written in other languages get only the guarantees provided by that language.
 
-Pony requires FFI parameters to be explicitly defined as Pony types in advance. The type system guarantees at compile-time that no code-paths exist that violate the definitions you provide on the Pony side of the FFI Interface.
+Pony requires FFI parameters to be explicitly declared as Pony types in advance. The type system provides a guarantee at compile-time that no code-paths exist that violate the definitions you provided on the Pony side of your FFI call.
 
-Unfortunately, Pony can't extend this guarantee to include the execution of external functions as it cannot possibly know if the code on the other side of the FFI Interface will break any of Pony's guarantees.
+This guarantee cannot be inclusive of foreign code as the compiler has no way to verify that it won't violate Pony's type system guarantees at runtime.
+
--- a/docs/types/at-a-glance.md
+++ b/docs/types/at-a-glance.md
@@ -37,6 +37,6 @@ Some of those will make sense right now. Some of them may not mean much to you y
 
 __If I use Pony's FFI to call code written in another language, does Pony magically make the same guarantees for the code I call?__ Sadly, no. Pony's type system can only guarantee code written in Pony. Code written in other languages gets only the guarantees provided by that language.
 
-Pony requires FFI parameters to be explicitly defined as Pony types in advance. This allows the typechecker to guarantee that no code-paths exist that would result in an FFI call that violated the top-level definition. This assurance is not available in many other languages.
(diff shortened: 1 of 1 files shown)


redvers:
When rendered that's quite a block of text. Going to go break that up a bit I think.

docs/types/at-a-glance.md
```diff
@@ -35,4 +35,9 @@ The Pony type system offers a lot of guarantees, even more than other statically
...
+__If I use Pony's FFI to call code written in another language, does Pony magically make the same guarantees for the code I call?__ Sadly, no. Pony's type system can only guarantee code written in Pony. Code written in other languages gets only the guarantees provided by that language.
+
+Pony requires FFI parameters be explicitly defined as Pony types in advance. This allows the typechecker to guarantee that no code-paths exist that would result in an FFI call that violated the top-level definition. This assurance is not available in many other languages.
+
+Unfortunately, As there is no way Pony can validate compatibility between your top-level definition and what the library expects Pony's guarantees cannot be extended.
```
EpicEric, line 42:
```suggestion
Unfortunately, as there is no way Pony can validate compatibility between your top-level definition and what the library expects, Pony's guarantees cannot be extended.
```

docs/types/at-a-glance.md
```diff
@@ -35,4 +35,9 @@ The Pony type system offers a lot of guarantees, even more than other statically
...
 
-__If I use Pony's FFI to call code written in another language, does Pony magically make the same guarantees for the code I call?__ Sadly, no. Pony has no way of determining if the FFI definition you provide is in the format your FFI library expects. Pony's type system does guarantee that data passed to your FFI calls match those type definitions you provide. This additional safety is not a guarantee. Pony's type system can only guarantee code written in Pony. Code written in other languages gets only the guarantees provided by that language.
+__If I use Pony's FFI to call code written in another language, does Pony magically make the same guarantees for the code I call?__ Sadly, no. Pony's type system can only guarantee code written in Pony. Code written in other languages gets only the guarantees provided by that language.
+
+Pony requires FFI parameters be explicitly defined as Pony types in advance. This allows the typechecker to guarantee that no code-paths exist that would result in an FFI call that violated the top-level definition. This assurance is not available in many other languages.
```
EpicEric, line 40:
```suggestion
Pony requires FFI parameters to be explicitly defined as Pony types in advance. This allows the typechecker to guarantee that no code-paths exist that would result in an FFI call that violated the top-level definition. This assurance is not available in many other languages.
```

EpicEric (changes requested):

ergl:
I'm not sure I agree completely with the text as it is written, at least the part about "This assurance is not available in many other languages". (Rust, for example, also forces you to specify a type specification for an FFI function).

I understand the paragraph as originally written as saying "Pony can't possibly know if the code on the other side of the FFI barrier will break any of the [guarantees](https://tutorial.ponylang.io/types/at-a-glance.html#what-guarantees-does-ponys-type-system-give-me) of the type system".

For example, a C library might use locks internally, and could cause your Pony program to deadlock if used incorrectly, or I could use the FFI to escape the capability guarantees that Pony offers:

```c
void* raw_cast(void* p)
{
  return p;
}
```

We can declare this function in a way that allows you to create a data race:

```pony
use @raw_cast[Any ref](obj: Any tag)

actor Counter
  var _count: U32 = 0
  be get(cb: {(U32): None} val) => cb(_count)
  be incr() => incr_fn()
  fun ref incr_fn() => _count = _count + 1

actor Main
  new create(env: Env) =>
    let c = Counter
    try
      let c' = @raw_cast(c) as Counter ref
      c.incr() // This line and the next one could be executed in any order
      c'.incr_fn()
      c.get({(v: U32) => env.out.print(v.string())}) // May print 1 or 2
    end
```

ergl:
I agree that we should still mention FFI declarations and the fact that Pony will type-check them, though!

redvers:
Right - and other things that have bitten me in the behind:

1. Who is responsible for memory management?  Is it pony or the API?  Who is responsible for freeing it?
2. Is the library you're binding to thread-safe?
3. Is that struct that you passed back and forward designed to be introspectable or opaque?

These are all things in the API docs which pony can't ascertain automatically hence why we can't guarantee.

We use FFI calls all over pony stdlib and we assert guarantees because many smart people have thought very hard and are willing to put their names on it.  We can't extend that confidence to anyone else :-)

So yes, there are all kinds of other aspects too - but since this was "at a glance" I tried to cut it down to the most obvious rusty-axe - Ensuring the 'shape' of the data is correct.

Pony can't ensure you're correct.  It can only ensure that you're consistent.

How to phrase these concepts in a few sentences is more challenging than I thought.

redvers:
> "Pony can't possibly know if the code on the other side of the FFI barrier will break any of the [guarantees]"

This is arguably a much simpler way to address it.  Something along the lines of:

"Pony requires FFI parameters to be explicitly defined as Pony types in advance. The type system guarantees at compile-time that no code-paths exist that violate the definitions you provide on the Pony side of the FFI Interface.

Unfortunately, Pony can't extend this guarantee to include the execution of external functions as it cannot possibly know if the code on the other side of the FFI Interface will break any of Pony's guarantees."

redvers:
> "Pony can't possibly know if the code on the other side of the FFI barrier will break any of the [guarantees]"

This is arguably a much simpler way to address it.  Something along the lines of:

"Pony requires FFI parameters to be explicitly defined as Pony types in advance. The type system guarantees at compile-time that no code-paths exist that violate the definitions you provide on the Pony side of the FFI Interface.

Unfortunately, Pony can't extend this guarantee to include the execution of external functions as it cannot possibly know if the code on the other side of the FFI Interface will break any of Pony's guarantees."

redvers:
This may be easier with a teeny diagram.

redvers:
```
          +      +-----------+
Pony      |      | Pony Code |
Guarantee |      +-----------+
Here      |  
          | +----------------------+
          | | Pony Type Definition |  <--+
          + +----------------------+     |
                                         |
               +---------------+         | Type model that
               | FFI Interface |         | maps Pony types
               +---------------+         | to equivalent
                                         | C types
          +        +-------+             |
          |        | C ABI |          <--+
No        |        +-------+
Guarantee |        
Available |    +--------------+
          |    | Library Code |
          +    +--------------+
```

redvers:
The additional safety the pony typechecker gives you is only as good as the model you provide it.  It can provide additional safety, but cannot provide guarantees.

(But the safety is a selling point)

ergl:
>"Pony requires FFI parameters to be explicitly defined as Pony types in advance. The type system guarantees at compile-time that no code-paths exist that violate the definitions you provide on the Pony side of the FFI Interface.
>
>Unfortunately, Pony can't extend this guarantee to include the execution of external functions as it cannot possibly know if the code on the other side of the FFI Interface will break any of Pony's guarantees."

Yes, I like this better. Do you want to add this to the PR? @redvers

docs/types/at-a-glance.md
```diff
@@ -35,4 +35,8 @@ The Pony type system offers a lot of guarantees, even more than other statically
 
 Some of those will make sense right now. Some of them may not mean much to you yet (like capabilities-security and causal messaging), but we'll get to those concepts later on.
 
-__If I use Pony's FFI to call code written in another language, does Pony magically make the same guarantees for the code I call?__ Sadly, no. Pony's type system guarantees only apply to code written in Pony. Code written in other languages gets only the guarantees provided by that language.
+__If I use Pony's FFI to call code written in another language, does Pony magically make the same guarantees for the code I call?__ Sadly, no. Pony's type system can only guarantee code written in Pony. Code written in other languages gets only the guarantees provided by that language.
```
rhagenson, line 38: Personal writing style comment here so feel free to ignore. I do not like the use of "magically" in this context as the question is asked from the perspective of the reader/learner so it seems dismissive to suggest they expect some "magic" -- although I certainly know I have expected magic before. Can you change this to be more open-ended and without the magic such as:
> __If I use Pony's FFI to call code written in another language, can Pony make the same guarantees for the code I call?__

rhagenson (changes requested):
Minor comments on FFI redundancy and a suggestion to drop the "magic" (even if we all hope for it at times).

docs/types/at-a-glance.md
redvers: It's not my writing style since I didn't write that section :-) ... but I'm happy to make that change if you think it's an improvement.

docs/types/at-a-glance.md
```diff
@@ -35,4 +35,8 @@ The Pony type system offers a lot of guarantees, even more than other statically
...
+
+Pony requires FFI parameters to be explicitly defined as Pony types in advance. The type system guarantees at compile-time that no code-paths exist that violate the definitions you provide on the Pony side of the FFI Interface.
+
+Unfortunately, Pony can't extend this guarantee to include the execution of external functions as it cannot possibly know if the code on the other side of the FFI Interface will break any of Pony's guarantees.
```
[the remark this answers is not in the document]
redvers: I know, and I'm also sensitive to the Department of Redundancy Department, but I didn't consider this to be an example of that which is why I left it in.

I considered that we were talking about the interface to the FFI system, not the FFI itself.

I can remove it if you wish, it reads a little clumsily to me but ymmv.

docs/types/at-a-glance.md
ergl: I agree with Ryan, I'd drop the "magic" term, the original sentence still works without it.

docs/types/at-a-glance.md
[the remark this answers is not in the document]
ergl: Perhaps something like the following?

> Unfortunately, the compiler can't verify that the execution of programs called through the FFI will not break Pony's type system guarantees.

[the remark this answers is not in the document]
rhagenson: I like Borja's rewrite to use "through" rather than "interface" to specify the "interface to the FFI system" as intended.

rhagenson:
The changes look good to me! I am on my phone right now and do not know how to change my "changes requested" to "approved" mobile. I do however approve of this now.

Red, thank you for thinking to add this! Giving people more reasons upfront to love Pony as much as we do.

rhagenson (approved):

ergl (approved):

EpicEric (approved):

PR #476, opened 2021-05-23
Remove lambda from keywords

ergl:
This is not used anymore now that the `{(a: Foo): Bar => ...}` syntax is used.


SeanTAllen:
Linting error is unrelated. Looks like someone merged something that wasn't passing linting. I'm fixing that in another commit.

PR #478, opened 2021-05-23
Add spellchecking

SeanTAllen:


SeanTAllen:
@rhagenson I think the biggest thing to review for this is... are there any words in our "additional dictionary" that make you go "o no, where is that used? maybe we should rewrite where ever that is".

one weird one...

"et al." had to be added as "et" and "al"

docs/types/at-a-glance.md
```diff
@@ -31,7 +31,7 @@ The Pony type system offers a lot of guarantees, even more than other statically
 * Your program will never deadlock.
 * Your code will always be capabilities-secure.
 * All message passing is
-[causal](https://courses.cs.vt.edu/~cs5204/fall00/causal.html). (Not ca**su**al!)
+[causal](https://courses.cs.vt.edu/~cs5204/fall00/causal.html). (Not `ca**su**al`!)
```
ergl, line 34: The backticks will show this word verbatim, perhaps it'd be better to set the entire word in bold?

SeanTAllen: That seems reasonable to me

SeanTAllen: I pushed that change.

rhagenson (commented):
Took a look at all of it and nothing I disagree with. I think changing some of the italics to be inline code style via backticks is odd in some places, but not distracting and certainly better than having to make one-off spellcheck exceptions.

"sequentiality" is the only entry in the wordlist that gave me pause; I had to look where it was used, but it makes sense in context.

rhagenson:
Having both "et" and "al" in the wordlist is odd, but not sure we can avoid needing them entirely. Any reference to a paper on Pony, such as any of the [Community Papers](https://www.ponylang.io/community/#papers) could lead to needing "et al." later on.

SeanTAllen:
@rhagenson is this good to merge?

rhagenson:
Yes, it looks good to me @SeanTAllen. Thank you for adding this action.

PR #488, opened 2022-01-19
Improve explanation of the 'as' keyword

### LINKED ISSUE #485: Update misleading `as` documentation
Reading the "As documentation"- https://tutorial.ponylang.io/pattern-matching/as.html#safely-converting-to-a-more-specific-type-casting I think it should be changed.

In particular it uses the words "casting" and "converting" neither of which are true for what "as" does.

I think it should be rewritten to talk about "as" as a shorthand for "match X type or error" at runtime. It never converts to another type, it never changes to another type.

There's some more nuance that needs to go into that related to how interfaces work and what people think it means in that example to "convert the interface" but I think this would be best explained via match not "casting".

I've seen a lot of people think that "as" changes the type of something which is not true, but reading the documentation in the tutorial, I can see why they think that.

After discussion in Zulip, it was decided that in general the format should be:

- every value has exactly one "true"/concrete type it is an instance of
- that concrete type can be subsumed inside of more general/abstract types that it is a subset/subtype of
- normal assignment/passing semantics let a value only move "down" into less specific / more general levels of abstraction (or stay at the same level it is)
- using as or match can let you move back "up" to more specific types by checking the type information at runtime

I personally think we should avoid using the terms "down" and "up" but at the moment, I don't have better terminology. If we do use "down" and "up" we should give examples that hopefully short-circuit the belief some people will have that subtyping in Pony is equivalent to inheritance in some other languages. I am worried that people will interpret "up" and "down" with "an inheritance mindset".

See https://ponylang.zulipchat.com/#narrow/stream/190368-tutorial/topic/As.20documentation for additional context.

niclash:
Closes #485

### DIFF
--- a/docs/pattern-matching/as.md
+++ b/docs/pattern-matching/as.md
@@ -4,18 +4,30 @@ The `as` operator in Pony has two related uses. First, it provides a safe way to
 
 ## Expressing a different type of an object
 
-As we know, each object instance has one or more types, and each alias is of one or more types. `as` allows an object alias to be expressed as one of the other types of the object which is not implied by the alias type. Simple example;
+In Pony, each object instance has a single concrete type and can fulfill one or more interfaces or traits. In short hand, we often refer to this as an instance having multiple types. `as` allows an object alias to be expressed as one of the other types of the object which is not implied by the alias type.
+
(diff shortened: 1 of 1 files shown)


SeanTAllen:
Im going to make some edits to this that @niclash and I discussed in zulip that he doesn't have time to make for a couple days.

SeanTAllen:
Closing because a force push made netlify angry. Opening a new PR.

PR #489, opened 2022-01-19
Improve `as` documentation

### LINKED ISSUE #485: Update misleading `as` documentation

SeanTAllen:
The previous documentation was misleading and could lead people
to believe that Pony does type-casting. Most of the heavy lifting
on this PR was done by Niclas Hedhman.

Closes #485


docs/pattern-matching/as.md
```diff
@@ -1,34 +1,48 @@
 # As Operator
 
-The `as` operator in Pony has two related uses. First, it provides a safe way to increase the specificity of an object's type (casting). Second, it gives the programmer a way to specify the type of the items in an array literal.
+The `as` operator in Pony has two related uses. First, it provides a safe way to increase the specificity of an object's type. Second, it gives the programmer a way to specify the type of the items in an array literal.
 
-## Safely converting to a more specific type (casting)
+## Expressing a different type of an object
 
-The `as` operator can be used to create a reference to an object with a more specific type than the given reference, if possible. This can be applied to types that are related through inheritance, as well as unions and intersections. This is done at runtime, and if it fails then an error is raised.
+In Pony, each object instance has a single concrete type and can fulfill one or more interfaces or traits. In short hand, we often refer to this as an instance having multiple types. `as` allows an object alias to be expressed as one of the other types of the object which is not implied by the alias type.
```
jemc, line 7: I had a lot of trouble reading through this paragraph, even given the context I already have about this feature. It doesn't map very well to the mental model I personally have, in that I don't think about it as "an instance having multiple types".

Let me take a shot at a different way of expressing it. Note that this is intended to replace both this paragraph and the one below it, but GitHub won't let me do a suggestion on the set pair of lines because it thinks that the set contains a deleted line.

```suggestion
In Pony, each object is an instance of a single concrete type, which is the most specific type for that object. But the object can also be held as part of a "wider" abstract type such as an interface, trait, or type union which the concrete type is said to be a subtype of.

`as` (like `match`) allows a program to check the runtime type of a abstract-typed value to see whether or not the object matches a given type which is more specific. If it doesn't match the more specific type, then a runtime `error` is raised. For example:
```

docs/pattern-matching/as.md
```diff
@@ -1,34 +1,48 @@
...
 
-The same thing can be done with interfaces, using `as` to create a reference to a more specific interface or class. Let's say, for example, that you have a library for doing things with furry, rodent-like creatures. It provides a `Critter` interface which programmers can then use to create specific types of critters.
+Note that the type requested as the `as` argument must exist as a type of the object instance, unlike C casting where one type can be forced into become another type. Type coercion is not possible in Pony, so one can not do `let value:F64 = F32(1.0) as F64`. `F32` and `F64` are both concrete types and each object can only have a single concrete type. Many concrete types do provide methods that allow you do convert them to another concrete type, for example, `F32(1.0).f64()` to convert an `F32` to an `F64` or `F32(1.0).string()` to convert to a string.
```
jemc, line 28:
```suggestion
Note that the type requested as the `as` argument must exist as a type of the object instance, unlike C casting where one type can be forced into become another type. Coercion from one concrete type to another is not possible using `as`, so one can not do `let value:F64 = F32(1.0) as F64`. `F32` and `F64` are both concrete types and each object can only have a single concrete type. Many concrete types do provide methods that allow you do convert them to another concrete type, for example, `F32(1.0).f64()` to convert an `F32` to an `F64` or `F32(1.0).string()` to convert to a string.
```

docs/pattern-matching/as.md
```diff
@@ -59,9 +73,56 @@ actor Main
...
+        | let critter: Critter =>
+          env.out.print(critter.wash())
+        else
+          error
+        end
+      end
+    end
+```
+
+Thinking of the `as` keyword as "an attempt to match that will error if not matched" is a good mental mode to have. If you don't care about handling the "not matched" case that causes an error when using `as`, you can rewrite an `as` to use match without an error like:
```
jemc, line 107: I don't know of "mental mode" or "mental model" was the intended term, but I think the latter is more natural here.

```suggestion
Thinking of the `as` keyword as "an attempt to match that will error if not matched" is a good mental model to have. If you don't care about handling the "not matched" case that causes an error when using `as`, you can rewrite an `as` to use match without an error like:
```

jemc (approved):

PR #490, opened 2022-01-20
Fix typos in `as` documentation

ergl:
The new version is a nice improvement! Just found out a couple of typos when reading it out loud.

### DIFF
--- a/docs/pattern-matching/as.md
+++ b/docs/pattern-matching/as.md
@@ -6,7 +6,7 @@ The `as` operator in Pony has two related uses. First, it provides a safe way to
 
 In Pony, each object is an instance of a single concrete type, which is the most specific type for that object. But the object can also be held as part of a "wider" abstract type such as an interface, trait, or type union which the concrete type is said to be a subtype of.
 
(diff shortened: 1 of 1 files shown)


docs/pattern-matching/as.md
```diff
@@ -23,11 +23,11 @@ In Pony, each object is an instance of a single concrete type, which is the most
...
+Note that the type requested as the `as` argument must exist as a type of the object instance, unlike C casting where one type can be forced to become another type. Coercion from one concrete type to another is not possible using `as`, so one can not do `let value:F64 = F32(1.0) as F64`. `F32` and `F64` are both concrete types and each object can only have a single concrete type. Many concrete types do provide methods that allow you to convert them to another concrete type, for example, `F32(1.0).f64()` to convert an `F32` to an `F64` or `F32(1.0).string()` to convert to a string.
 
-In addition to using to using `as` with a union of disjoint types, we can also express an intersected type of the object, meaning the object has a type that the alias we have for the object is not directly related to the type we want to express. Example;
```
ergl, line 30: I had touble parsing what the second part of this paragraph is supposed to say. I tried rewording it but couldn't think of something sensible.

PR #494, opened 2022-01-30
Elaborate on trait vs interface differences

SeanTAllen:
This commit provides a rewrite of the traits and interfaces documentation.
The goals are:

- Be more nuanced in the differences between trait and interface

In particular, acknowledge that interfaces can be used for both
structural and nominal subtyping.

- Explain the differences between trait and interface

The previous content was focused almost entirely on nominal vs
structural subtyping and didn't give folks who aren't experienced
with both examples to demonstrate the advantages of each approach
as well as the different affordances they provide.

### DIFF
--- a/docs/types/traits-and-interfaces.md
+++ b/docs/types/traits-and-interfaces.md
@@ -211,4 +211,4 @@ class Compactor
     end
 ```
 
-The flexibility of interface has allowed us to define a type `Compactable` that we can use to allow our `Compactor` to accept a variety of data types including `Array`, `Map`, and `String` from the standard library.
+The flexibility of `interface` has allowed us to define a type `Compactable` that we can use to allow our `Compactor` to accept a variety of data types including `Array`, `Map`, and `String` from the standard library.
--- a/docs/types/traits-and-interfaces.md
+++ b/docs/types/traits-and-interfaces.md
@@ -98,7 +98,7 @@ Here, we have a class `Larry` that has a `name` function with the same signature
 
 __Wait, why not?__ Because `Larry` doesn't say it `is Named`. Remember, traits are __nominal__: a type that wants to provide a trait has to explicitly declare that it does. And `Larry` doesn't.
 
-You can also do nominal subtyping using the keyword `interface`. __Interfaces__ in Pony are primarily used for structural subtyping. However, interfaces, like traits can have default method implementations and in order to use default method implementations, an interface must be used in a nominal fashion. For example:
+You can also do nominal subtyping using the keyword `interface`. __Interfaces__ in Pony are primarily used for structural subtyping. Like traits, interfaces can also have default method implementations, but in order to use default method implementations, an interface must be used in a nominal fashion. For example:
 
 ```pony
 interface HasName
--- a/docs/types/traits-and-interfaces.md
+++ b/docs/types/traits-and-interfaces.md
@@ -143,7 +143,7 @@ interface Foo
   fun ref _set(i: USize, value: U8): U8
 ```
 
-In the code above, the interface `Foo` allows access to the private `_set` method of String and allows for changing `sailor` to `failor` or it would anyway, if interfaces were allowed to have private methods.
+In the code above, the interface `Foo` allows access to the private `_set` method of `String` and allows for changing `sailor` to `failor` or it would anyway, if interfaces were allowed to have private methods.
 
 ### Open world enumerations
 
--- a/docs/types/traits-and-interfaces.md
+++ b/docs/types/traits-and-interfaces.md
@@ -12,9 +12,45 @@ If you've done object-oriented programming before, you may have seen a lot of di
 
 The core idea is that you have a type that declares it has a relationship to some category type. In Java, for example, a __class__ (a concrete type) can __implement__ an __interface__ (a category type). In Java, this means the class is now in the category that the interface represents. The compiler will check that the class actually provides everything it needs to.
 
-## Traits: nominal subtyping
+In Pony, nominal subtyping is done via the keyword `is`. `is` declares at the point of declaration that an object has a relationship to a category type. For example, to use nominal subtyping to declare that the class `Name` provides `Stringable`, you'd do:
 
-Pony has nominal subtyping, using __traits__. A __trait__ looks a bit like a __class__, but it uses the keyword `trait` and it can't have any fields.
+```pony
+class Name is Stringable
+```
+
+## Structural subtyping
+
+There's another kind of subtyping, where the name doesn't matter. It's called __structural subtyping__, which means that it's all about how a type is built, and nothing to do with names.
+
+A concrete type is a member of a structural category if it happens to have all the needed elements, no matter what it happens to be called.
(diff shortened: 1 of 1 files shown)


docs/types/traits-and-interfaces.md
```diff
@@ -12,9 +12,45 @@ If you've done object-oriented programming before, you may have seen a lot of di
...
+class Name is Stringable
+```
+
+## Structural subtyping
+
+There's another kind of subtyping, where the name doesn't matter. It's called __structural subtyping__, which means that it's all about how a type is built, and nothing to do with names.
+
+A concrete type is a member of a structural category if it happens to have all the needed elements, no matter what it happens to be called.
+
+Structural typing is very similar to [duck typing](https://en.wikipedia.org/wiki/Duck_typing) from dynamic programming languages except that, type compatibility is determined at compile time rather than at run time. If you've used Go, you'll recognise that Go interfaces are structural types.
```
ergl, line 27:
```suggestion
Structural typing is very similar to [duck typing](https://en.wikipedia.org/wiki/Duck_typing) from dynamic programming languages, except that type compatibility is determined at compile time rather than at run time. If you've used Go, you'll recognise that Go interfaces are structural types.
```

docs/types/traits-and-interfaces.md
```diff
@@ -12,9 +12,45 @@ If you've done object-oriented programming before, you may have seen a lot of di
...
+
+You do not declare structural relationships ahead of time, instead it done by checking if a given concrete type can fulfill the required interface. For example, in the code below, we have the interface `Stringable` from the standard library. Anything can be used as a `Stringable` so long as it provides the method `fun string(): String iso^`. In our example below, `ExecveError` provides the `Stringable` interface and can be used anywhere a `Stringable` is needed. Because `Stringable` is a structural type, `ExecveError` doesn't have to declare a relationship to `Stringable`, it simply has that relationship because it has "the same shape".
```
ergl, line 29:
```suggestion
You do not declare structural relationships ahead of time, instead it is done by checking if a given concrete type can fulfill the required interface. For example, in the code below, we have the interface `Stringable` from the standard library. Anything can be used as a `Stringable` so long as it provides the method `fun string(): String iso^`. In our example below, `ExecveError` provides the `Stringable` interface and can be used anywhere a `Stringable` is needed. Because `Stringable` is a structural type, `ExecveError` doesn't have to declare a relationship to `Stringable`, it simply has that relationship because it has "the same shape".
```

docs/types/traits-and-interfaces.md
```diff
@@ -62,17 +98,23 @@ Here, we have a class `Larry` that has a `name` function with the same signature
 
 __Wait, why not?__ Because `Larry` doesn't say it `is Named`. Remember, traits are __nominal__: a type that wants to provide a trait has to explicitly declare that it does. And `Larry` doesn't.
 
-## Structural subtyping
+You can also do nominal subtyping using the keyword `interface`. __Interfaces__ in Pony are primarily used for structural subtyping. However, interfaces, like traits can have default method implementations and in order to use default method implementations, an interface must be used in a nominal fashion. For example:
```
ergl, line 101: I would move this paragraph and its example to the "Structural subtyping" section, since both focus on `interfaces`. I've also suggested some changes to the last part of the paragraph, since I found the word a bit confusing.

```suggestion
You can also do nominal subtyping using the keyword `interface`. __Interfaces__ in Pony are primarily used for structural subtyping. Like traits, interfaces can also have default method implementations, but in order to use default method implementations, an interface must be used in a nominal fashion. For example:
```

docs/types/traits-and-interfaces.md
```diff
@@ -81,6 +123,92 @@ interface HasName
...
+    let x: String ref = "sailor".string()
+    let y: Foo = x
+    y._set(0, 'f')
+    env.out.print("Hello, " + x)
+
+interface Foo
+  fun ref _set(i: USize, value: U8): U8
+```
+
+In the code above, the interface `Foo` allows access to the private `_set` method of String and allows for changing `sailor` to `failor` or it would anyway, if interfaces were allowed to have private methods.
```
ergl, line 146:
```suggestion
In the code above, the interface `Foo` allows access to the private `_set` method of `String` and allows for changing `sailor` to `failor` or it would anyway, if interfaces were allowed to have private methods.
```

docs/types/traits-and-interfaces.md
```diff
@@ -81,6 +123,92 @@ interface HasName
...
+Traits allow you to create "open world enumerations" that others can participate in. For example:
+
+```pony
+trait Color
+
+primitive Red is Color
+primitive Blue is Color
+```
+
+Here we are using a trait to provide a category of things, `Color` that any things can opt into by declaring itself to be a `Color`. The creates an "open world" of enumerations that you can't do using the more traditional Pony approach using type unions.
```
ergl, line 159:
```suggestion
Here we are using a trait to provide a category of things, `Color`, that any other types can opt into by declaring itself to be a `Color`. This creates an "open world" of enumerations that you can't do using the more traditional Pony approach using type unions.
```

docs/types/traits-and-interfaces.md
```diff
@@ -81,6 +123,92 @@ interface HasName
...
+
+```pony
+interface Color
+  fun is_color(): None
+
+primitive Red
+  fun is_color(): None => None
+```
+
+Here we are using structural typing to create a collection of things that are in the category `Color` by providing a method that "marks" being a member of the category; `is_color`.
```
ergl, line 186:
```suggestion
Here we are using structural typing to create a collection of things that are in the category `Color` by providing a method that "marks" being a member of the category: `is_color`.
```

docs/types/traits-and-interfaces.md
```diff
@@ -81,6 +123,92 @@ interface HasName
...
+    _threshold = threshold
+
+  fun ref try_compacting(thing: Compactable) =>
+    if thing.size() > _threshold then
+      thing.compact()
+    end
+```
 
-__Should I use traits or interfaces in my own code?__ Both! Interfaces are more flexible, so if you're not sure what you want, use an interface. But traits are a powerful tool as well: they stop _accidental subtyping_.
+The flexibility of interface has allowed us to define a type `Compactable` that we can use to allow our `Compactor` to accept a variety of data types including `Array`, `Map`, and `String` from the standard library.
```
ergl, line 214:
```suggestion
The flexibility of `interface` has allowed us to define a type `Compactable` that we can use to allow our `Compactor` to accept a variety of data types including `Array`, `Map`, and `String` from the standard library.
```

ergl (commented):
I think this looks good overall, have left some minor comments and suggestions.

docs/types/traits-and-interfaces.md
SeanTAllen: I disagree with the moving. I think it is best to introduce both concepts then cover them particularly as the interface keyword allows for a bit of nominal as well as structural.

ergl (approved):

PR #496, opened 2022-02-08
Add documentation for the nosupertype annotation

### LINKED ISSUE #492: `nosupertype` annotation is missing from tutorial
See https://github.com/ponylang/rfcs/blob/main/text/0054-subtyping-exclusion.md for information needed to add.

SeanTAllen:
Closes #492

### DIFF
--- a/docs/appendices/annotations.md
+++ b/docs/appendices/annotations.md
@@ -111,7 +111,7 @@ The above code won't compile because you could supply `Empty ref`. Doing so resu
 By adding `nosupertype` to the definition of `Empty`, we declare that `Empty` is not a subtype of `Any` and thereby allow the code to compile as there is no longer an unsafe match.
 
 ```pony
-class \nosuperclass\ Empty
+class \nosupertype\ Empty
 
 class Foo
   fun foo[A: Any](a: (A | Empty val)) =>
--- a/docs/appendices/annotations.md
+++ b/docs/appendices/annotations.md
@@ -89,3 +89,35 @@ class \nodoc\Foo
    We don't want this class and its methods to appear in generated documentation
    """
 ```
+
+### `nosupertype`
+
+Recognised on objects(`actor`, `class`, `primitive`, `struct`). A type annotated with `nosupertype` will not be a subtype of any other type (except _), even if the type structurally provides an interface. If a `nosupertype` type has a provides list, a compiler error is reported. As a result, a `nosupertype` type is excluded from both nominal and structural subtyping.
+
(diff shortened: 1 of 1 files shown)


docs/appendices/annotations.md
```diff
@@ -89,3 +89,35 @@ class \nodoc\Foo
...
+    | let a': A => None
+    end
+```
+
+The above code won't compile because you could supply `Empty ref`. Doing so results in a compiler error about an unsafe match because we would need to distinguish between `Empty val` and `Empty ref` at runtime.
+
+By adding `nosupertype` to the definition of `Empty`, we declare that `Empty` is not a subtype of `Any` and thereby allow the code to compile as there is no longer an unsafe match.
+
+```pony
+class \nosuperclass\ Empty
```
jemc, line 114: Wrong annotation name?
```suggestion
class \nosupertype\ Empty
```

docs/appendices/annotations.md
```diff
@@ -89,3 +89,35 @@ class \nodoc\Foo
...
+
+class Foo
+  fun foo[A: Any](a: (A | Empty val)) =>
+    match consume a
+    | let a': A => None
+    end
+```
+
+`nosuperclass` is particularly valuable when constructing generic classes like collections that need a marker class to describe "lack of an item".
```
jemc, line 123: Wrong annotation name?
```suggestion
`nosupertype` is particularly valuable when constructing generic classes like collections that need a marker class to describe "lack of an item".
```

jemc (commented):
I totally forgot this feature even existed.

I don't even remember discussing, voting on, or approving the RFC.

SeanTAllen:
It was part of a 2 parter from Benoit. Only half got implemented.

docs/appendices/annotations.md
```diff
@@ -89,3 +89,35 @@ class \nodoc\Foo
    We don't want this class and its methods to appear in generated documentation
    """
 ```
+
+### `nosupertype`
+
+Recognised on objects(`actor`, `class`, `primitive`, `struct`). A type annotated with `nosupertype` will not be a subtype of any other type (except _), even if the type structurally provides an interface. If a `nosupertype` type has a provides list, a compiler error is reported. As a result, a `nosupertype` type is excluded from both nominal and structural subtyping.
```
ergl, line 95: What's the meaning of `_` in this context?

SeanTAllen: I lifted that directly from the RFC. I need to go look and see again what specifically `_` is in the type system.

PR #503, opened 2022-03-15
Remove extraneous ‘that’

dusty-phillips:
Looks like this was once “we’ll cover that” and was supposed to be switched to “that we’ll cover”, but… well, that’s that.


rhagenson (approved):
That's that!

PR #506, opened 2022-03-19
Add new default reference types page

dusty-phillips:
One thing I struggle with as a new Pony rider is figuring out what the
default capability is if one isn't specified.

Judging by https://ponylang.zulipchat.com/#narrow/stream/189985-beginner-help/topic/Default.20reference.20capabilities
I am not the only one.

This new tutorial page was partially informed by the defaults claimed in
the rest of the tutorial, but a lot of it came by way of trial and
compiler error. I tried to come up with as many different "default"
scenarios as I could and implemented them to discover what happened.

Some notes for reviewers:

* I'm not certain I got all of these correct. I have something like two
  hours of Pony experience under my belt. I'm guessing there a massive
  holes and nuances I missed.
* I'm not certain I found all the places default capabilities are used.
* I may not have used the terminology exactly correctly.

That said, I think the layout and format are mostly ok, so this should
be pretty easy to build on.

I've run mkdocs locally and confirmed that the nav link works and the markdown file renders nicely.

### DIFF
--- a/docs/reference-capabilities/default-capabilities.md
+++ b/docs/reference-capabilities/default-capabilities.md
@@ -1,0 +1,234 @@
+# Default Reference Types
+
+We've mentioned throughout this tutorial what the default reference capabilities are for types and classes. However, there's a good chance those defaults passed over your head while you were focused on understanding capabilities themselves.
+
+First, there is nothing saying you have to use default capabilities. Especially as you start writing your first Pony programs, you'll probably find it very educational to explicitly specify the capabilities for all your classes, actors, functions, parameters and fields. However, it's not _writing_ code that makes default capabilities confusing; it's _reading_ that code. Hopefully keeping this summary open will help you navigate pony source code until you've become used to the defaults!
+
+## Reference type summary
+
+We go into detail of the default types below, but here's a table summarizing them:
+
+| Item                    | Default Capability | Explanation                                                             |
+| ----------------------- | ------------------ | ----------------------------------------------------------------------- |
+| primitive               | val                | Primitives are immutable and can never be anything but `val`            |
+| actor                   | tag                | Actors can neither be read nor written and are always `tag`             |
+| class                   | ref                | Instances of classes default to ref type if unspecified                 |
+| fields                  | default for type   | Whatever was specified on the _target_ class                            |
+| variables               | default for type   | Whatever was specified on the _target_ class                            |
+| parameter               | default for type   | Whatever was specified on the _target_ class                            |
+| constructor (primitive) | val                | Only classes can have different constructor capabilities                |
+| constructor (actor)     | tag                | Only classes can have different constructor capabilities                |
+| constructor (class)     | ref                | Only classes can have different constructor capabilities                |
+| receiver (function)     | box                | Functions cannot mutate class state by default (use `func ref`)         |
+| receiver (behaviour)    | ref                | Actors can mutate their state from inside a behaviour                   |
+| recover (immutable)     | val                | An immutable value (box or val) is recovered to `val` by default        |
+| recover (mutable)       | iso                | An immutable value (box or val) is recovered to `val` by default        |
+| literal (object)        | ref                | object literals default to same capabilities as their non-literal types |
+| literal (actor)         | tag                | object literals default to same capabilities as their non-literal types |
+| literal (primitive)     | val                | object literals default to same capabilities as their non-literal types |
+| lambda (no captures)    | val                | lambda capabilities depend on whether they capture their environment    |
+| lambda (captures)       | ref                | lambda capabilities depend on whether they capture their environment    |
+
+## Primitive default capabilities
+
+Let's start with primitives because they're easy. All primitives have `val` capabilities. They are immutable, they are read only, there is only one of each primitive, and they are accessible from all actors. So `val`.
+
+In fact, defining a primitive as `primitive val MyPrimitive` is a syntax error. Primitives are so `val` that you can't even tell them they are `val`.
+
+## Actor default capabilities
+
+Similar to primitives, actors always have a `tag` default capability. Like primitives, it is not possible to specify a different default capability for tags.
+
+```
+actor Actor
+```
+
+has `tag` capabilities
+
+```
+actor ref Actor
+```
+
+is illegal
+
+## Class default capabilities
+
+By default, a class has `ref` capability. Therefore, these two class definitions are equivalent:
+
+```
+class ref RefClass
+    ...
+
+class RefClass
+    ...
+```
+
+But whether specified by default or explicitly as `class ref` (or `class val` or any other capability) what does it actually mean? A class or actor's default capability only applies when you _define_ a field or local type. This can happen either as `let` or `var`, or when specifying a parameter on a function or behaviour.
+
+## Field, Variable, and Parameter capabilities
+
+For example, `String` is defined in the standard library as `class val String`. So it's default capability is the immutable shareable `val` capability. That means that all the `String`s in the following actor definition could be replaced with `String val`:
+
+```
+actor UsesStrings
+    let s1: String
+    var s2: String
+
+    new create(s1': String, s2': String) =>
+        s1 = s1'
+        s2 = s2'
+
+    be do_thing(s3: String) =>
+        None
+
+```
+
+However `StringBytes` is defined as `class ref`, so all the `StringBytes`' in the following class definition are equivalent to `StringBytes ref`:
+
+```
+class UsesStringBytes
+    let s1: StringBytes
+    var s2: StringBytes
+
+    new create(s1': StringBytes, s2': StringBytes) =>
+        s1 = s1'
+        s2 = s2'
+
+    fun do_thing(s3: StringBytes) =>
+        None
+```
+
+Notably, however, the default capability on a class **does not** influence the constructors for that class.
+
+## Constructor capabilities
+
+Let's start with the easy defaults first: A primitive constructor _always_ returns an object with `val` capabilities, and an actor constructor _always_ returns an object with `tag` capabilities. In fact, it is forbidden to put a capablity after `new` when defining a primitive or actor.
+
+Classes are a little more flexible. The capability returned by a `new` constructor default to a `ref` if you are constructing a `class`. This is true regardless of what the default capability for the class is, but you have the option to specify a different capability by explicitly typing the constructor as e.g `new var create` or `new iso create_iso`.
+
+Consider the following class:
+
+```
+class val MyClass
+    let foo: String
+
+    new create_ref(foo': String) =>
+        foo = foo'
+
+    new val create_val(foo': String) =>
+        foo = foo'
+```
+
+This class has two constructors. The first one, `create_ref` returns a `MyClass` with `ref` capabilities because `ref` is always default for class constructors. The second one `create_val` explicitly returns a `MyClass` with `val` capabilities.
+
+_Tip_: Instead of using the default `ref` type for constructors, consider using `iso`. Recall that constructors, like `consume` return their values as an ephemeral type. That means that a constructor typed as `new iso` actually returns a `iso^`. And `iso^` has the terrific property of being convertible to any other type! Consider:
+
+```
+class MyIsoClass
+    let foo: String
+
+    new iso create(foo': String) =>
+        foo = foo'
+```
+
+This class can be construted as a ref or val or whatever else:
+
+```
+a_ref: MyIsoClass ref = MyIsoClass.create()
+a_val: MyIsoClass val = MyIsoClass.create()
+```
+
+In fact, this is what happens if you do not specify a constructor at all. The hidden constructor for this class also returns an `iso^`:
+
+```
+class AllDefaults
+    var foo: String = "hello"
+```
+
+You can construct instances of this class with a specific capability by specifying the type of the result:
+
+```
+let a = AllDefaults // a is iso
+let b: AllDefaults ref = AllDefaults // b is ref
+let c: AllDefaults val = AllDefaults // c is val
+```
+
+Constructors create objects, but what about when you access functions on those objects?
+
+## Default receiver capabilities
+
+Recall that when you call a function or behaviour, it always implicitly gets a reference to the `this` object.
+
+The default receiver capability for a function is `box`, as shown in `get_foo` below:
+
+```
+class MyClass
+    var foo: String = ""
+
+    fun get_foo() => foo
+
+    fun ref set_foo(foo': String) => foo = foo'
+```
+
+Notably, you can't mutate a `box`, so mutable methods must be defined as `fun ref` or `fun trn`.
+
+The _only_ behavior capability for an actor is `ref`. You cannot specify a `be val` for example, or even `be ref`. This may be unintuitive since the _caller_ only has a tag to the actor, but the actor is totally free to reference its own fields from the behaviour.
+
+## Default recover capabilities
+
+If you are recovering a value, it will default to `val` if the value is immutable (`val` or `box`), and `iso` if it is mutable (`ref` or `trn`):
+
+```
+var i_am_iso = recover MyClass.create_ref("abc") end
+var i_am_val = recover MyClass.create_box("def") end
+```
+
+You can, of course, change the capability using `recover ref` or similar.
+
+## Default object literal capabilities
+
+The default capabilities on an object literal depend on whether the object literal specifies a primitive, actor or class, and use the same defaults as the accompanying type:
+
+```
+let myObject = object
+    let s: String = "hello"
+end
+```
+
+has `ref` capabilities (it is an object of an anonymous class).
+
+```
+let myActor = object
+    be apply() => None
+end
+```
+
+has tag capabilities.
+
+let myPrimitive = object
+fun apply() => None
+end
+
+```
+
+has `val` capabilities.
+```
+
+## Default lambda capabilities
+
+Recall that a lambda creates an object literal with an `apply` function. That underlying object has `ref` capabilities if it captures any variables. If it does not capture variables, it defaults to `val` capabilities.
+
+```
+var myLambdaVal = {(s: String): String => "my Lambda" + s}
+var myLambdaRef = {(s: String): String => myLambdaVal("here") + s}
+```
+
+`myLambdaVal` has `val` capabilities because it doesn't capture anything. `myLambdaRef` defaults to `ref` capabilities because it captures `myLambdaVal`.
+
+You can specify different capabilities after the lambda:
+
+```
+var myOtherLambdaRef = {(): String => "hello"} ref
+```
+
+is a `ref` even though with no captures it would have defaulted to `val` capabilities.
--- a/docs/reference-capabilities/default-reference-types.md
+++ b/docs/reference-capabilities/default-reference-types.md
@@ -1,0 +1,234 @@
+# Default Reference Capabilities on Types
+
+We've mentioned throughout this tutorial what the default reference capabilities are for types and classes. However, there's a good chance those defaults passed over your head while you were focused on understanding capabilities themselves.
+
+First, there is nothing saying you have to use default capabilities. Especially as you start writing your first Pony programs, you'll probably find it very educational to explicitly specify the capabilities for all your classes, actors, functions, parameters and fields. However, it's not _writing_ code that makes default capabilities confusing; it's _reading_ that code. Hopefully keeping this summary open will help you navigate pony source code until you've become used to the defaults!
+
+## Reference type summary
+
+We go into detail of the default types below, but here's a table summarizing them:
+
+| Item                    | Default Capability | Explanation                                                             |
+| ----------------------- | ------------------ | ----------------------------------------------------------------------- |
+| primitive               | val                | Primitives are immutable and can never be anything but `val`            |
+| actor                   | tag                | Actors can neither be read nor written and are always `tag`             |
+| class                   | ref                | Instances of classes default to ref type if unspecified                 |
+| fields                  | default for type   | Whatever was specified on the _target_ class                            |
+| variables               | default for type   | Whatever was specified on the _target_ class                            |
+| parameter               | default for type   | Whatever was specified on the _target_ class                            |
+| constructor (primitive) | val                | Only classes can have different constructor capabilities                |
+| constructor (actor)     | tag                | Only classes can have different constructor capabilities                |
+| constructor (class)     | ref                | Only classes can have different constructor capabilities                |
+| receiver (function)     | box                | Functions cannot mutate class state by default (use `func ref`)         |
+| receiver (behaviour)    | ref                | Actors can mutate their state from inside a behaviour                   |
+| recover (immutable)     | val                | An immutable value (box or val) is recovered to `val` by default        |
+| recover (mutable)       | iso                | An immutable value (box or val) is recovered to `val` by default        |
+| literal (object)        | ref                | object literals default to same capabilities as their non-literal types |
+| literal (actor)         | tag                | object literals default to same capabilities as their non-literal types |
+| literal (primitive)     | val                | object literals default to same capabilities as their non-literal types |
+| lambda (no captures)    | val                | lambda capabilities depend on whether they capture their environment    |
+| lambda (captures)       | ref                | lambda capabilities depend on whether they capture their environment    |
+
+## Primitive default capabilities
+
+Let's start with primitives because they're easy. All primitives have `val` capabilities. They are immutable, they are read only, there is only one of each primitive, and they are accessible from all actors. So `val`.
+
+In fact, defining a primitive as `primitive val MyPrimitive` is a syntax error. Primitives are so `val` that you can't even tell them they are `val`.
+
+## Actor default capabilities
+
+Similar to primitives, actors always have a `tag` default capability. Like primitives, it is not possible to specify a different default capability for tags.
+
+```
+actor Actor
+```
+
+has `tag` capabilities
+
+```
+actor ref Actor
+```
+
+is illegal
+
+## Class default capabilities
+
+By default, a class has `ref` capability. Therefore, these two class definitions are equivalent:
+
+```
+class ref RefClass
+    ...
+
+class RefClass
+    ...
+```
+
+But whether specified by default or explicitly as `class ref` (or `class val` or any other capability) what does it actually mean? A class or actor's default capability only applies when you _define_ a field or local type. This can happen either as `let` or `var`, or when specifying a parameter on a function or behaviour.
+
+## Field, Variable, and Parameter capabilities
+
+For example, `String` is defined in the standard library as `class val String`. So it's default capability is the immutable shareable `val` capability. That means that all the `String`s in the following actor definition could be replaced with `String val`:
+
+```
+actor UsesStrings
+    let s1: String
+    var s2: String
+
+    new create(s1': String, s2': String) =>
+        s1 = s1'
+        s2 = s2'
+
+    be do_thing(s3: String) =>
+        None
+
+```
+
+However `StringBytes` is defined as `class ref`, so all the `StringBytes`' in the following class definition are equivalent to `StringBytes ref`:
+
+```
+class UsesStringBytes
+    let s1: StringBytes
+    var s2: StringBytes
+
+    new create(s1': StringBytes, s2': StringBytes) =>
+        s1 = s1'
+        s2 = s2'
+
+    fun do_thing(s3: StringBytes) =>
+        None
+```
+
+Notably, however, the default capability on a class **does not** influence the constructors for that class.
+
+## Constructor capabilities
+
+Let's start with the easy defaults first: A primitive constructor _always_ returns an object with `val` capabilities, and an actor constructor _always_ returns an object with `tag` capabilities. In fact, it is forbidden to put a capablity after `new` when defining a primitive or actor.
+
+Classes are a little more flexible. The capability returned by a `new` constructor default to a `ref` if you are constructing a `class`. This is true regardless of what the default capability for the class is, but you have the option to specify a different capability by explicitly typing the constructor as e.g `new var create` or `new iso create_iso`.
+
+Consider the following class:
+
+```
+class val MyClass
+    let foo: String
+
+    new create_ref(foo': String) =>
+        foo = foo'
+
+    new val create_val(foo': String) =>
+        foo = foo'
+```
+
+This class has two constructors. The first one, `create_ref` returns a `MyClass` with `ref` capabilities because `ref` is always default for class constructors. The second one `create_val` explicitly returns a `MyClass` with `val` capabilities.
+
+_Tip_: Instead of using the default `ref` type for constructors, consider using `iso`. Recall that constructors, like `consume` return their values as an ephemeral type. That means that a constructor typed as `new iso` actually returns a `iso^`. And `iso^` has the terrific property of being convertible to any other type! Consider:
+
+```
+class MyIsoClass
+    let foo: String
+
+    new iso create(foo': String) =>
+        foo = foo'
+```
+
+This class can be construted as a ref or val or whatever else:
+
+```
+a_ref: MyIsoClass ref = MyIsoClass.create()
+a_val: MyIsoClass val = MyIsoClass.create()
+```
+
+In fact, this is what happens if you do not specify a constructor at all. The hidden constructor for this class also returns an `iso^`:
+
+```
+class AllDefaults
+    var foo: String = "hello"
+```
+
+You can construct instances of this class with a specific capability by specifying the type of the result:
+
+```
+let a = AllDefaults // a is iso
+let b: AllDefaults ref = AllDefaults // b is ref
+let c: AllDefaults val = AllDefaults // c is val
+```
+
+Constructors create objects, but what about when you access functions on those objects?
+
+## Default receiver capabilities
+
+Recall that when you call a function or behaviour, it always implicitly gets a reference to the `this` object.
+
+The default receiver capability for a function is `box`, as shown in `get_foo` below:
+
+```
+class MyClass
+    var foo: String = ""
+
+    fun get_foo() => foo
+
+    fun ref set_foo(foo': String) => foo = foo'
+```
+
+Notably, you can't mutate a `box`, so mutable methods must be defined as `fun ref` or `fun trn`.
+
+The _only_ behavior capability for an actor is `ref`. You cannot specify a `be val` for example, or even `be ref`. This may be unintuitive since the _caller_ only has a tag to the actor, but the actor is totally free to reference its own fields from the behaviour.
+
+## Default recover capabilities
+
+If you are recovering a value, it will default to `val` if the value is immutable (`val` or `box`), and `iso` if it is mutable (`ref` or `trn`):
+
+```
+var i_am_iso = recover MyClass.create_ref("abc") end
+var i_am_val = recover MyClass.create_box("def") end
+```
+
+You can, of course, change the capability using `recover ref` or similar.
+
+## Default object literal capabilities
+
+The default capabilities on an object literal depend on whether the object literal specifies a primitive, actor or class, and use the same defaults as the accompanying type:
+
+```
+let myObject = object
+    let s: String = "hello"
+end
+```
+
+has `ref` capabilities (it is an object of an anonymous class).
+
+```
+let myActor = object
+    be apply() => None
+end
+```
+
+has tag capabilities.
+
+let myPrimitive = object
+fun apply() => None
+end
+
+```
+
+has `val` capabilities.
+```
+
+## Default lambda capabilities
+
+Recall that a lambda creates an object literal with an `apply` function. That underlying object has `ref` capabilities if it captures any variables. If it does not capture variables, it defaults to `val` capabilities.
+
+```
+var myLambdaVal = {(s: String): String => "my Lambda" + s}
+var myLambdaRef = {(s: String): String => myLambdaVal("here") + s}
+```
+
+`myLambdaVal` has `val` capabilities because it doesn't capture anything. `myLambdaRef` defaults to `ref` capabilities because it captures `myLambdaVal`.
+
+You can specify different capabilities after the lambda:
+
+```
+var myOtherLambdaRef = {(): String => "hello"} ref
+```
+
+is a `ref` even though with no captures it would have defaulted to `val` capabilities.
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -112,6 +112,7 @@ nav:
       - Combining Capabilities: "reference-capabilities/combining-capabilities.md"
       - Arrow Types aka Viewpoints: "reference-capabilities/arrow-types.md"
       - Reference Capability Matrix: "reference-capabilities/capability-matrix.md"
+      - Default Reference Types: "reference-capabilities/default-capabilities.md"
   - Object Capabilities:
       - Overview: "object-capabilities/index.md"
       - Object Capabilities: "object-capabilities/object-capabilities.md"


rhagenson:
@dusty-phillips Thank you for the contribution. I will review it in about an hour after I grab some lunch.

A quick look on my phone now and I will be making some suggestions in my review.

docs/reference-capabilities/default-capabilities.md
```diff
@@ -0,0 +1,233 @@
...
+| actor                   | `tag`              | Actors can neither be read nor written and are always `tag`             |
+| class                   | `ref`              | Instances of classes default to ref type if unspecified                 |
+| fields                  | default for type   | Whatever was specified on the _target_ class                            |
+| variables               | default for type   | Whatever was specified on the _target_ class                            |
+| parameter               | default for type   | Whatever was specified on the _target_ class                            |
+| constructor (primitive) | `val`              | Only classes can have different constructor capabilities                |
+| constructor (actor)     | `tag`              | Only classes can have different constructor capabilities                |
+| constructor (class)     | `ref`              | Only classes can have different constructor capabilities                |
+| receiver (function)     | `box`              | Functions cannot mutate class state by default (use `func ref`)         |
```
ergl, line 22:
```suggestion
| receiver (function)     | `box`              | Functions cannot mutate class state by default (use `fun ref`)         |
```

docs/reference-capabilities/default-capabilities.md
```diff
@@ -0,0 +1,233 @@
...
+
+First, there is nothing saying you have to use default capabilities. Especially as you start writing your first Pony programs, you'll probably find it very educational to explicitly specify the capabilities for all your classes, actors, functions, parameters and fields. However, it's not _writing_ code that makes default capabilities confusing; it's _reading_ that code. Hopefully keeping this summary open will help you navigate pony source code until you've become used to the defaults!
+
+## Reference type summary
+
+We go into detail of the default types below, but here's a table summarizing them:
+
+| Item                    | Default Capability | Explanation                                                             |
+| ----------------------- | ------------------ | ----------------------------------------------------------------------- |
+| primitive               | `val`              | Primitives are immutable and can never be anything but `val`            |
```
jasoncarr0, line 13: While primitives are immutable, reference capabilities apply to references. One could in theory have a `P tag` reference to some primitive `P`.

Perhaps: are always constructed as `val`

docs/reference-capabilities/default-capabilities.md
```diff
@@ -0,0 +1,233 @@
...
+| actor                   | `tag`              | Actors can neither be read nor written and are always `tag`             |
```
jasoncarr0, line 14: Again, each actor has a `ref` reference to itself. Perhaps: Actors are always seen by other actors as `tag`.

docs/reference-capabilities/default-capabilities.md
```diff
@@ -0,0 +1,233 @@
...
+| variables               | default for type   | Whatever was specified on the _target_ class                            |
+| parameter               | default for type   | Whatever was specified on the _target_ class                            |
+| constructor (primitive) | `val`              | Only classes can have different constructor capabilities                |
+| constructor (actor)     | `tag`              | Only classes can have different constructor capabilities                |
+| constructor (class)     | `ref`              | Only classes can have different constructor capabilities                |
+| receiver (function)     | `box`              | Functions cannot mutate class state by default (use `func ref`)         |
+| receiver (behaviour)    | `ref`              | Actors can mutate their state from inside a behaviour                   |
+| recover (immutable)     | `val`              | An immutable value (box or val) is recovered to `val` by default        |
+| recover (mutable)       | `iso`              | An immutable value (box or val) is recovered to `val` by default        |
```
jasoncarr0, line 25: Typo here?

docs/reference-capabilities/default-capabilities.md
```diff
@@ -0,0 +1,233 @@
...
+| literal (object)        | `ref`              | object literals default to same capabilities as their non-literal types |
+| literal (actor)         | `tag`              | object literals default to same capabilities as their non-literal types |
+| literal (primitive)     | `val`              | object literals default to same capabilities as their non-literal types |
+| lambda (no captures)    | `val`              | lambda capabilities depend on whether they capture their environment    |
```
jasoncarr0, line 29: no *possibly mutable* captures

docs/reference-capabilities/default-capabilities.md
```diff
@@ -0,0 +1,233 @@
...
+| lambda (captures)       | `ref`              | lambda capabilities depend on whether they capture their environment    |
+
+## Primitive default capabilities
+
+Let's start with primitives because they're easy. All primitives have `val` capabilities. They are immutable, they are read only, there is only one of each primitive, and they are accessible from all actors. So `val`.
```
rhagenson, line 34:
```suggestion
Let's start with primitives because they're stateless. All primitives have `val` capabilities. Primitives are immutable, they are read-only, there is only one of each primitive, and it is accessible from all actors -- read-only and shareable means the default is `val`.
```

I try to resist calling anything "easy" or even "simple" as it can have the effect of subtly stating "if you don't understand this, then you cannot understand it at all" which is not true and especially bad in a tutorial setting where whether the reader understands it is on us as the writers.

docs/reference-capabilities/default-capabilities.md
```diff
@@ -0,0 +1,233 @@
+# Default Reference Types
+
+We've mentioned throughout this tutorial what the default reference capabilities are for types and classes. However, there's a good chance those defaults passed over your head while you were focused on understanding capabilities themselves.
+
+First, there is nothing saying you have to use default capabilities. Especially as you start writing your first Pony programs, you'll probably find it very educational to explicitly specify the capabilities for all your classes, actors, functions, parameters and fields. However, it's not _writing_ code that makes default capabilities confusing; it's _reading_ that code. Hopefully keeping this summary open will help you navigate pony source code until you've become used to the defaults!
```
SeanTAllen, line 5:
```suggestion
First, there is nothing saying you have to use default capabilities. Especially as you start writing your first Pony programs, you'll might find it educational to explicitly specify the capabilities for all your classes, actors, functions, parameters and fields. However, it's not _writing_ code that makes default capabilities confusing; it's _reading_ that code. Hopefully keeping this summary open will help you navigate pony source code until you've become used to the defaults!
```

SeanTAllen, line 5: I don't like the assumption of difficulty here. Much as we shouldn't assume things are easy for people, this assumes hard for people. I'd like to this rewritten to be to not make assumption about if the reader is finding the topic easy or hard but rather address that if they are finding it hard, don't worry, they aren't alone in that. Some people catch on with reference capabilities easily, others struggle with different parts.

SeanTAllen, line 5: This says "First," but there's no "Second". Were you intending to add another point? If not, I'd like to remove the "First," but I'm not sure if in the process of writing if you had some additional points you were thinking of making and then didn't while working out other parts.

docs/reference-capabilities/default-capabilities.md
```diff
@@ -0,0 +1,233 @@
...
+| literal (actor)         | `tag`              | object literals default to same capabilities as their non-literal types |
+| literal (primitive)     | `val`              | object literals default to same capabilities as their non-literal types |
+| lambda (no captures)    | `val`              | lambda capabilities depend on whether they capture their environment    |
+| lambda (captures)       | `ref`              | lambda capabilities depend on whether they capture their environment    |
+
+## Primitive default capabilities
+
+Let's start with primitives because they're easy. All primitives have `val` capabilities. They are immutable, they are read only, there is only one of each primitive, and they are accessible from all actors. So `val`.
+
+In fact, defining a primitive as `primitive val MyPrimitive` is a syntax error. Primitives are so `val` that you can't even tell them they are `val`.
```
rhagenson, line 36:
```suggestion
In fact, defining a primitive as `primitive val MyPrimitive` is a syntax error.
```

I like the second sentence as it brings in some personality to the Tutorial, but nonetheless I think it is not needed.

docs/reference-capabilities/default-capabilities.md
```diff
@@ -0,0 +1,233 @@
...
+
+## Actor default capabilities
+
+Similar to primitives, actors always have a `tag` default capability. Like primitives, it is not possible to specify a different default capability for tags.
```
rhagenson, line 40:
```suggestion
Actors always have a default `tag` capability. Like primitives, attempting to override the default reference capability is a syntax error.
```

docs/reference-capabilities/default-capabilities.md
```diff
@@ -0,0 +1,233 @@
+# Default Reference Types
+
+We've mentioned throughout this tutorial what the default reference capabilities are for types and classes. However, there's a good chance those defaults passed over your head while you were focused on understanding capabilities themselves.
+
+First, there is nothing saying you have to use default capabilities. Especially as you start writing your first Pony programs, you'll probably find it very educational to explicitly specify the capabilities for all your classes, actors, functions, parameters and fields. However, it's not _writing_ code that makes default capabilities confusing; it's _reading_ that code. Hopefully keeping this summary open will help you navigate pony source code until you've become used to the defaults!
+
+## Reference type summary
```
SeanTAllen, line 7: So I think it should be assumed that someone arrives on this page from a search and additional information should be here as well. Before diving into a table, I think there should be a recapping of what a "default capability" is. Something like showing...

```pony
fun foo(): String iso^ =>
  "hello world".clone()
```

and then show the

```pony
fun box foo(): String iso^ =>
  "hello world".clone()
```

docs/reference-capabilities/default-capabilities.md
```diff
@@ -0,0 +1,233 @@
...
+actor Actor
+```
+
+has `tag` capabilities
+
+```pony
+actor ref Actor
+```
+
+is illegal
```
rhagenson, line 52:
```suggestion
is illegal. This is because a actor `ref` would not be concurrency safe -- it would allow directly reading an actor's internal state as opposed to requiring message passing to share state.
```

docs/reference-capabilities/default-capabilities.md
```diff
@@ -0,0 +1,233 @@
...
+
+We've mentioned throughout this tutorial what the default reference capabilities are for types and classes. However, there's a good chance those defaults passed over your head while you were focused on understanding capabilities themselves.
+
+First, there is nothing saying you have to use default capabilities. Especially as you start writing your first Pony programs, you'll probably find it very educational to explicitly specify the capabilities for all your classes, actors, functions, parameters and fields. However, it's not _writing_ code that makes default capabilities confusing; it's _reading_ that code. Hopefully keeping this summary open will help you navigate pony source code until you've become used to the defaults!
+
+## Reference type summary
+
+We go into detail of the default types below, but here's a table summarizing them:
+
+| Item                    | Default Capability | Explanation                                                             |
```
SeanTAllen, line 11: I find numerous things in this table confusing. I understand it is trying to be expansive but I think it ends up being rather confusing in some places, like listing defaults for things and then saying "well but only this other type can have different capabilities".

I think a distinction needs to be made between "default" as in "you can override" and "this is what it is" and only discuss the former and not mix in the latter in the same section.

docs/reference-capabilities/default-capabilities.md
```diff
@@ -0,0 +1,233 @@
...
+
+```pony
+class ref RefClass
+    ...
+
+class RefClass
+    ...
+```
+
+But whether specified by default or explicitly as `class ref` (or `class val` or any other capability) what does it actually mean? A class or actor's default capability only applies when you _define_ a field or local type. This can happen either as `let` or `var`, or when specifying a parameter on a function or behaviour.
```
rhagenson, line 66:
```suggestion
But whether specified by default or explicitly as `class ref` (or `class val` or any other capability) what does it actually mean? A class or actor's default capability only applies when you _define_ a field or local type. This can happen when using `let` or `var`, or when specifying a parameter on a function or behaviour.
```

rhagenson: The "either" made me think it referred to the "or" between let and var, rather than the continuation after the comma.

docs/reference-capabilities/default-capabilities.md
```diff
@@ -0,0 +1,233 @@
...
+
+## Field, Variable, and Parameter capabilities
+
+For example, `String` is defined in the standard library as `class val String`. So it's default capability is the immutable shareable `val` capability. That means that all the `String`s in the following actor definition could be replaced with `String val`:
+
+```pony
+actor UsesStrings
```
SeanTAllen, line 73: indentation should be two spaces not four.

docs/reference-capabilities/default-capabilities.md
```diff
@@ -0,0 +1,233 @@
...
+    let s1: String
+    var s2: String
+
+    new create(s1': String, s2': String) =>
+        s1 = s1'
+        s2 = s2'
+
+    be do_thing(s3: String) =>
+        None
+
```
SeanTAllen, line 83:
```suggestion
```

docs/reference-capabilities/default-capabilities.md
```diff
@@ -0,0 +1,233 @@
...
+class ref RefClass
+    ...
+
+class RefClass
+    ...
+```
+
+But whether specified by default or explicitly as `class ref` (or `class val` or any other capability) what does it actually mean? A class or actor's default capability only applies when you _define_ a field or local type. This can happen either as `let` or `var`, or when specifying a parameter on a function or behaviour.
+
+## Field, Variable, and Parameter capabilities
```
rhagenson, line 68:
```suggestion
## Field, variable, and parameter capabilities
```

docs/reference-capabilities/default-capabilities.md
```diff
@@ -0,0 +1,233 @@
...
+
+For example, `String` is defined in the standard library as `class val String`. So it's default capability is the immutable shareable `val` capability. That means that all the `String`s in the following actor definition could be replaced with `String val`:
```
rhagenson, line 70:
```suggestion
For example, `String` is defined in the standard library as `class val String`. So its default capability is the immutable, shareable `val` capability. That means that all the `String`s in the following actor definition could be replaced with `String val`:
```

docs/reference-capabilities/default-capabilities.md
```diff
@@ -0,0 +1,233 @@
...
+    var s2: String
+
+    new create(s1': String, s2': String) =>
+        s1 = s1'
+        s2 = s2'
+
+    be do_thing(s3: String) =>
+        None
+
+```
```
rhagenson, line 84:
```suggestion
```pony
actor UsesStrings
    let s1: String
    var s2: String

    new create(s1': String, s2': String) =>
        s1 = s1'
        s2 = s2'

    be do_thing(s3: String) =>
        None
```
```

File: docs/reference-capabilities/default-capabilities.md [hunk 19]
```diff
@@ -0,0 +1,233 @@
...
+
+    new create(s1': StringBytes, s2': StringBytes) =>
+        s1 = s1'
+        s2 = s2'
+
+    fun do_thing(s3: StringBytes) =>
+        None
+```
+
+Notably, however, the default capability on a class **does not** influence the constructors for that class.
```
rhagenson, line 101: What do you mean here?

docs/reference-capabilities/default-capabilities.md
```diff
@@ -0,0 +1,233 @@
...
+```
+
+This class can be constructed as a ref or val or whatever else:
+
+```pony
+a_ref: MyIsoClass ref = MyIsoClass.create()
+a_val: MyIsoClass val = MyIsoClass.create()
+```
+
+In fact, this is what happens if you do not specify a constructor at all. The hidden constructor for this class also returns an `iso^`:
```
SeanTAllen, line 141: I don't like the term "hidden" here. I don't think calling constructor that wasn't explicitly written by the author as "hidden" is accurate. Nothing was done to hide it. Every object must have a constructor. A default will be provided if the author doesn't supply one.

docs/reference-capabilities/default-capabilities.md
```diff
@@ -0,0 +1,233 @@
...
+
+    fun do_thing(s3: StringBytes) =>
+        None
+```
+
+Notably, however, the default capability on a class **does not** influence the constructors for that class.
+
+## Constructor capabilities
+
+Let's start with the easy defaults first: A primitive constructor _always_ returns an object with `val` capabilities, and an actor constructor _always_ returns an object with `tag` capabilities. In fact, it is forbidden to put a capability after `new` when defining a primitive or actor.
```
rhagenson, line 105:
```suggestion
Let's start with the required defaults first: a primitive constructor _always_ returns an object with `val` capabilities, and an actor constructor _always_ returns an object with `tag` capabilities. In fact, it is forbidden to put a capability after `new` when defining a primitive or actor.
```

Not sure "required" is the right word, but I mean to remove the "easy" while implying their requirement for primitive and actor to be their specified defaults.

SeanTAllen:
I left some comments. I have some additional possible concerns but they might be alleviate if the overall content is revisited in response to some of the comments already left.

docs/reference-capabilities/default-capabilities.md
rhagenson: Do you mean the ability to override the default on a given constructor?

docs/reference-capabilities/default-capabilities.md
SeanTAllen: I think a lot of these explanations while well intended can be misleading and this table overall would be better without the explanations. I'd like to see an attempt made to make the overall content of this page work without the explanation field of the table that I think is ambiguous in places and potentially misleading.

docs/reference-capabilities/default-capabilities.md
```diff
@@ -0,0 +1,233 @@
...
+
+Classes are a little more flexible. The capability returned by a `new` constructor default to a `ref` if you are constructing a `class`. This is true regardless of what the default capability for the class is, but you have the option to specify a different capability by explicitly typing the constructor as e.g `new var create` or `new iso create_iso`.
```
rhagenson, line 107:
```suggestion
Classes are more flexible. The capability returned by a constructor defaults to `ref` if you are constructing a `class`. This is true regardless of what the default capability for the class is, but you have the option to specify a different capability by explicitly typing the constructor (e.g.,  `new iso create_iso`).
```

docs/reference-capabilities/default-capabilities.md
```diff
@@ -0,0 +1,233 @@
...
+    let foo: String
+
+    new create_ref(foo': String) =>
+        foo = foo'
+
+    new val create_val(foo': String) =>
+        foo = foo'
+```
+
+This class has two constructors. The first one, `create_ref` returns a `MyClass` with `ref` capabilities because `ref` is always default for class constructors. The second one `create_val` explicitly returns a `MyClass` with `val` capabilities.
```
rhagenson, line 122:
```suggestion
This class has two constructors. The first one, `create_ref` returns a `MyClass` with `ref` capability because `ref` is always default for class constructors. The second one `create_val` explicitly returns a `MyClass` with `val` capability.
```

"Capabilities" refers to the collection of refcaps in my mind, while "capability" refers to a single one of the refcaps -- is this a consistent thought, or do more people thing of refcaps in terms of their specific denials?

docs/reference-capabilities/default-capabilities.md
```diff
@@ -0,0 +1,233 @@
...
+
+_Tip_: Instead of using the default `ref` type for constructors, consider using `iso`. Recall that constructors, like `consume` return their values as an ephemeral type. That means that a constructor typed as `new iso` actually returns a `iso^`. And `iso^` has the terrific property of being convertible to any other type! Consider:
```
rhagenson, line 124:
```suggestion
_Tip_: Instead of using the default `ref` capability for constructors, consider using `iso`. Recall that constructors, like `consume` return their values as an ephemeral type. That means that a constructor typed as `new iso` actually returns a `iso^`. And `iso^` has the terrific property of being convertible to any other property! Consider:
```

docs/reference-capabilities/default-capabilities.md
```diff
@@ -0,0 +1,233 @@
...
+
+```pony
+class MyIsoClass
+    let foo: String
+
+    new iso create(foo': String) =>
+        foo = foo'
+```
+
+This class can be constructed as a ref or val or whatever else:
```
rhagenson, line 134:
```suggestion
This class can be constructed as a `ref` or `val` or other capability:
```

docs/reference-capabilities/default-capabilities.md
rhagenson, line 141: Not sure I understand this. Can you elaborate?

docs/reference-capabilities/default-capabilities.md
```diff
@@ -0,0 +1,233 @@
...
+```
+
+In fact, this is what happens if you do not specify a constructor at all. The hidden constructor for this class also returns an `iso^`:
+
+```pony
+class AllDefaults
+    var foo: String = "hello"
+```
+
+You can construct instances of this class with a specific capability by specifying the type of the result:
```
rhagenson, line 148: Does not including a `new create()` here cause it to be `iso` by default? If so, that is news to me and may be why I was confused just above.

SeanTAllen:
Thanks for taking a stab at this. There's a good amount of stuff I want to discuss in relation to it, but I hope that isn't seen as a discouragement. I think that default capabilities and a focused discussion of just them is a great idea for inclusion in the tutorial.

docs/reference-capabilities/default-capabilities.md
```diff
@@ -0,0 +1,233 @@
...
+let a = AllDefaults // a is iso
+let b: AllDefaults ref = AllDefaults // b is ref
+let c: AllDefaults val = AllDefaults // c is val
+```
+
+Constructors create objects, but what about when you access functions on those objects?
+
+## Default receiver capabilities
+
+Recall that when you call a function or behaviour, it always implicitly gets a reference to the `this` object.
```
rhagenson, line 160:
```suggestion
Recall that when you call a function or behaviour, it always implicitly gets a reference to itself in the form of `this`.
```

docs/reference-capabilities/default-capabilities.md
```diff
@@ -0,0 +1,233 @@
...
+    var foo: String = ""
+
+    fun get_foo() => foo
+
+    fun ref set_foo(foo': String) => foo = foo'
+```
+
+Notably, you can't mutate a `box`, so mutable methods must be defined as `fun ref` or `fun trn`.
+
+The _only_ behavior capability for an actor is `ref`. You cannot specify a `be val` for example, or even `be ref`. This may be counter-intuitive since the _caller_ only has a tag to the actor, but the actor is totally free to reference its own fields from the behaviour.
```
rhagenson, line 175: Is this true? I have never thought to check this so if it is then I just learned something! I cannot think of a reason for it not to be true besides the implementer wanting to create some form of "pass through" behavior which is statically guaranteed not to mutate internal state, but would be allowed to send messages (perhaps will copies of internal state).

docs/reference-capabilities/default-capabilities.md
```diff
@@ -0,0 +1,233 @@
...
+## Default recover capabilities
+
+If you are recovering a value, it will default to `val` if the value is immutable (`val` or `box`), and `iso` if it is mutable (`ref` or `trn`):
+
+```pony
+var i_am_iso = recover MyClass.create_ref("abc") end
+var i_am_val = recover MyClass.create_box("def") end
+```
+
+You can, of course, change the capability using `recover ref` or similar.
```
rhagenson, line 186:
```suggestion
You can still change the capability using `recover ref` or similar.
```

docs/reference-capabilities/default-capabilities.md
```diff
@@ -0,0 +1,233 @@
...
+
+## Default object literal capabilities
+
+The default capabilities on an object literal depend on whether the object literal specifies a primitive, actor or class, and use the same defaults as the accompanying type:
```
rhagenson, line 190:
```suggestion
The default capabilities on an object literal depend on whether the object literal specifies a primitive, actor, or class, and use the same defaults as the accompanying type:
```

Oxford commas rule!

docs/reference-capabilities/default-capabilities.md
```diff
@@ -0,0 +1,233 @@
...
+
+has `ref` capabilities (it is an object of an anonymous class).
+
+```pony
+let myActor = object
+    be apply() => None
+end
+```
+
+has tag capabilities.
```
rhagenson, line 206:
```suggestion
has `tag` capabilities as it is an anonymous actor due to defining at least one behavior.
```

docs/reference-capabilities/default-capabilities.md
```diff
@@ -0,0 +1,233 @@
...
+```
+
+You can, of course, change the capability using `recover ref` or similar.
+
+## Default object literal capabilities
+
+The default capabilities on an object literal depend on whether the object literal specifies a primitive, actor or class, and use the same defaults as the accompanying type:
+
+```pony
+let myObject = object
```
rhagenson, line 193:
```suggestion
let myClass = object
```

docs/reference-capabilities/default-capabilities.md
```diff
@@ -0,0 +1,233 @@
...
+    let s: String = "hello"
+end
+```
+
+has `ref` capabilities (it is an object of an anonymous class).
```
rhagenson, line 198:
```suggestion
has `ref` capabilities as it is an anonymous class due to having functions, no behaviors, and maintaining internal state.
```

docs/reference-capabilities/default-capabilities.md
```diff
@@ -0,0 +1,233 @@
...
+end
+```
+
+has tag capabilities.
+
+```pony
+let myPrimitive = object
+fun apply() => None
+end
+```
```
rhagenson, line 212:
```suggestion
```pony
let myPrimitive = object
    fun apply() => None
end
```
```

File: docs/reference-capabilities/default-capabilities.md [hunk 35]
```diff
@@ -0,0 +1,233 @@
...
+
+has `val` capabilities.
```
rhagenson, line 214:
```suggestion
has `val` capabilities as it is an anonymous primitive due to having functions, no behaviors, and no internal state.
```

docs/reference-capabilities/default-capabilities.md
```diff
@@ -0,0 +1,233 @@
...
+
+## Default lambda capabilities
+
+Recall that a lambda creates an object literal with an `apply` function. That underlying object has `ref` capabilities if it captures any variables. If it does not capture variables, it defaults to `val` capabilities.
```
rhagenson, line 218:
```suggestion
Recall that a lambda creates an object literal with an `apply` function. That underlying object has `ref` capability if it captures any variables. If it does not capture variables, it defaults to `val` capability.
```

docs/reference-capabilities/default-capabilities.md
```diff
@@ -0,0 +1,233 @@
...
+
+```pony
+var myLambdaVal = {(s: String): String => "my Lambda" + s}
+var myLambdaRef = {(s: String): String => myLambdaVal("here") + s}
+```
+
+`myLambdaVal` has `val` capabilities because it doesn't capture anything. `myLambdaRef` defaults to `ref` capabilities because it captures `myLambdaVal`.
+
+You can specify different capabilities after the lambda:
```
rhagenson, line 227:
```suggestion
You can specify a different capability after the lambda:
```

docs/reference-capabilities/default-capabilities.md
```diff
@@ -0,0 +1,233 @@
...
+
+```pony
+var myOtherLambdaRef = {(): String => "hello"} ref
+```
+
+is a `ref` even though with no captures it would have defaulted to `val` capabilities.
```
rhagenson, line 233:
```suggestion
is a `ref` even though with no captures it would have defaulted to `val` capability.
```

rhagenson (changes requested):
Excellent addition! I am fairly certain I learned many things about defaults just by reviewing this which means I am thoroughly impressed at the depth of your contribution. I made a number of suggested changes and asked questions throughout for clarity in language and mechanics.

docs/reference-capabilities/default-capabilities.md
dusty-phillips: That’s been tripping me up all along. Perhaps two separate tables.

dusty-phillips:
Thank you so much. I love how fast and thorough you all are! I’ll make the suggested changes, make sure I understand them, and do a solid editing pass tomorrow before requesting rereview.

docs/reference-capabilities/default-capabilities.md
SeanTAllen: +1 to avoiding easy and simple. Simple does have a definition that is useful in technical writing, sadly that meaning is rarely understood so I try to avoid "simple" like I avoid "easy".

docs/reference-capabilities/default-capabilities.md
SeanTAllen: So I think there's a misunderstanding here on what a "default capability on a class" is.

It's not a "default on the class" it is the default capability for the type if the type appears elsewhere in the source without a capability.

So...

`class val String` means that if I say: `let x: String` it means `let x: String val`.

Which has nothing to do with the capability of the instance created when calling a given constructor.

docs/reference-capabilities/default-capabilities.md
SeanTAllen: `new create()` is not `new iso create()` it is `new ref create()`.

docs/reference-capabilities/default-capabilities.md
SeanTAllen: No it's not true. I understand the idea trying to be expressed, but it isn't true.

This is a misunderstanding of what is happening with viewpoint adaptation. `be` is not `ref`. If it was, then you couldn't send a message to an actor because all you have is a `tag`.

There's other points of  "kind of right, but not quite right" in the explanations here that I was saving to address later.

docs/reference-capabilities/default-capabilities.md
SeanTAllen: I disagree with this change. It's an object, not a class.

However, I also disagree with "object literal specifies a primitive, actor or class". An object literal is an object literal. primitives, actors, and classes are all objects. Saying an object literal specifies one while well meanging isn't correct and that explanation could be problematic in some contexts.

docs/reference-capabilities/default-capabilities.md
jasoncarr0: What about removing the clone and returning `String val` so that you can show the default capability for String?

docs/reference-capabilities/default-capabilities.md
SeanTAllen: This isn't correct. There's no such thing as an anonymous primitive. Is this a primitive?

```pony
class Foo
  fun apply(): U64 =>
    3
```

No, its a class. But it has functions, no behaviors, and no internal state.

SeanTAllen:
@dusty-phillips if you are available at 3 eastern (US), then I think that we could do some group editing/working on this during office hours some time. It might help with discussion of some finer points and larger conversations about structure etc.

docs/reference-capabilities/default-capabilities.md
SeanTAllen: Ah so `String` as the return type in the first and `String val` in the second?

SeanTAllen:
@dusty-phillips I'm glad you appreciate the thorough. Many folks can find it overwhelming and deflating.

docs/reference-capabilities/default-capabilities.md
jasoncarr0: Exactly, two birds with one stone that way

docs/reference-capabilities/default-capabilities.md
ergl: I think the correct term here would be "to any other capability"

docs/reference-capabilities/default-capabilities.md
dusty-phillips: Yeah, that was the confusion I was trying to resolve here. I originally expected there to be some correlation between the return type on the constructor and spent quite a bit of trial and error before it clicked that the constructor return type has nothing to do with the default on the class. It's one of those things that is too obvious to an experienced Pony developer and a new user needs to be told only once. Possibly it should go at the point class default reference capabilities are first introduced instead of here.

docs/reference-capabilities/default-capabilities.md
dusty-phillips: I didn't make a conscious decision about it, and I assume that what you are expecting is consistent with the other documentation. In my mind, a reference capability type may have multiple capabilites in the sense that it may be capable of multiple things (specifically, "read", "write", and "share"). But much more likely that I just wasn't paying attention to the existing docs.

docs/reference-capabilities/default-capabilities.md
dusty-phillips: ah, yes; I never thought about how defaults are affected by viewpoints. Most of this came from "instantiate a thing and see what the compiler thinks it is if I assign it to something else (as opposed to "I know what's going on" ;-) ). My eyes may have glossed over a `->` somewhere.

docs/reference-capabilities/default-capabilities.md
dusty-phillips: Can we solve this with judicious use of the word "instance"? Is it pedantically correct to say "an object literal with behaviours defined is an instance of an actor type"?

dusty-phillips:
My god, Github's "commit suggestion" flow is not designed for bulk suggestions. I will clean the history up by squashing all the suggestions and force pushing. Then I'll address the currently unresolved feedback and do a full editing pass.

> @dusty-phillips if you are available at 3 eastern (US), then I think that we could do some group editing/working on this during office hours some time. It might help with discussion of some finer points and larger conversations about structure etc.

Would love to, but I've such a bad cough that I'm not really capable of human-human interaction right now. For now I'll get this into as good a state as I can with my current knowledge before that time. If you like, I can hand it off to you for something between "final sanitizing" and "complete rewrite" (I won't be offended!) at that time. On the other hand, if you've got better things to do with your time, I'll keep poking it until I understand what I'm talking about.

> @dusty-phillips I'm glad you appreciate the thorough. Many folks can find it overwhelming and deflating.

Quality technical reviewers are much harder to find than willing writers. I've been both, so I know how to appreciate it!

docs/reference-capabilities/default-capabilities.md
dusty-phillips: Changed it as this and added a note about `tag` to the text below.

docs/reference-capabilities/default-capabilities.md
dusty-phillips: I'm leaving the explanations in place for now, but as placeholders. I'm not sure if they can be salvaged or not, but there isn't a lot of sense in updating them until the content is correct.

docs/reference-capabilities/default-capabilities.md
dusty-phillips: `yyp` fail. 😳

docs/reference-capabilities/default-capabilities.md
dusty-phillips: I was parrotting the vibes from the intro to https://tutorial.ponylang.io/reference-capabilities/capability-matrix.html here. Now that you mention it, I actually find reference capabilities relatively intuitive, and this document was my only confusion. I'll tone it down.

dusty-phillips: I meant it more as "Before we begin", but if I phrase it that way the whole paragraph falls apart. I'll stitch it together differently when I push changes.

docs/reference-capabilities/default-capabilities.md
SeanTAllen: That's misleading. Try turning your thinking around from an object literal to "actor is a keyword that creates an object with certain qualities". Ditto for class and primitive. It's a subtle but i think important distinction.

SeanTAllen: An object that has behaviors will have a message queue. We often refer to these as "actor" and we have a keyword for creating these objects. 

I don't think your statement is wrong but I do worry that it could be misleading.

SeanTAllen:
A concern... it easier for many when first learning references capabilities to think of them as what they allow but that's a more limited model than what it going on. Reference capabilities are formulated as what they deny and this distinction becomes important in "advanced cases". 

There's language in here that suggests that ref caps are formulated as what they allow. 

Eventually, before this gets merged, I'd want to do a review to make sure that doesn't end up being misleading.

docs/reference-capabilities/default-capabilities.md
dusty-phillips: I created one with a class as well to get even more birds.

docs/reference-capabilities/default-capabilities.md
dusty-phillips: I split it into two tables, but I still don't love it. I don't have any specific ideas, so I'll leave this as an open question.

docs/reference-capabilities/default-capabilities.md
dusty-phillips: I didn't like it either, but my brain was stuck. Changed to "auto-generated".

docs/reference-capabilities/default-capabilities.md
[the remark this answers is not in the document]
dusty-phillips: I think my new intro solves this.

docs/reference-capabilities/default-capabilities.md
```diff
@@ -0,0 +1,233 @@
...
+| primitive               | `val`              | Primitives are immutable and can never be anything but `val`            |
+| actor                   | `tag`              | Actors can neither be read nor written and are always `tag`             |
+| class                   | `ref`              | Instances of classes default to ref type if unspecified                 |
+| fields                  | default for type   | Whatever was specified on the _target_ class                            |
+| variables               | default for type   | Whatever was specified on the _target_ class                            |
+| parameter               | default for type   | Whatever was specified on the _target_ class                            |
+| constructor (primitive) | `val`              | Only classes can have different constructor capabilities                |
+| constructor (actor)     | `tag`              | Only classes can have different constructor capabilities                |
+| constructor (class)     | `ref`              | Only classes can have different constructor capabilities                |
```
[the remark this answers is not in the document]
dusty-phillips: I'm leaving this open as part of the "the table explanations column needs work" discussion.

docs/reference-capabilities/default-capabilities.md
dusty-phillips: I think github made this suggestion weird and it was supposed to just be deleting the extra newline. If I missed another change in it, let me know.

docs/reference-capabilities/default-capabilities.md
dusty-phillips: Looking at this again, I realized that the chief purpose of this sentence was actually as a transition to the next section. I massaged it a bit, but it might still be a bit stiff.

docs/reference-capabilities/default-capabilities.md
dusty-phillips: Changed to `capability`, but I will leave this comment unresolved. Mainly, I need confirmation that this is a good tip.

docs/reference-capabilities/default-capabilities.md
dusty-phillips: Does it make sense after I changed 'hidden' to 'auto-generated'?

docs/reference-capabilities/default-capabilities.md
dusty-phillips: oops, so it is. I changed this to:

```
let a = AllDefaults // a is ref
let b: AllDefaults iso = AllDefaults // b is iso
let c: AllDefaults val = AllDefaults // c is val
```

However, I'm leaving the discussion open, as I realized the entire constructors section is overlooking the fact that constructors return ephemeral types.

docs/reference-capabilities/default-capabilities.md
dusty-phillips: I'm not touching this paragraph. I can see that it' wrong, but I think @SeanTAllen will need to set it right.

docs/reference-capabilities/default-capabilities.md
dusty-phillips: I've reworked this a little, but now it feels more accurate, but also clumsy.

docs/reference-capabilities/default-capabilities.md
dusty-phillips: Note: I changed my mind on the editing pass and removed the Explanation column.

docs/reference-capabilities/default-capabilities.md
dusty-phillips: Ok, I'm confused. Consider this code:

```pony
let a = AllDefaults
let a2: AllDefaults val = a
```

This is obviously a compile-time error. And here it is:

```
Error:
/Users/dustyphillips/Desktop/Code/pony-experiments/referee/referee.pony:66:33: right side must be a subtype of left side
        let a2: AllDefaults val = a
                                ^
    Info:
    /Users/dustyphillips/Desktop/Code/pony-experiments/referee/referee.pony:157:1: AllDefaults iso is not a subtype of AllDefaults val^: iso is not a subcap of val^
    class AllDefaults
    ^
    /Users/dustyphillips/Desktop/Code/pony-experiments/referee/referee.pony:157:1: this would be possible if the subcap were more ephemeral. Perhaps you meant to consume a variable here
    class AllDefaults
    ^
```

The reason I thought `a` in this case has `iso` capabilities is that phrase `iso is not a subcap of val^`.

SeanTAllen: Sorry, what are you confused by?

dusty-phillips:
I've made some changes in response to the excellent reviews. At this point, I think the document captures *my* understanding better than it captures reality.

@SeanTAllen in lieu of a shared editing session (I desperately need a nap), my inclination is to hand ownership of the PR to you for further editing. I'd learn a lot from whatever commits you added to correct/improve it.

However, I don't like dumping half-finished trash on people, so feel no compunction about declining said ownership. If that is the case, I'll probably wander off and do some more Pony development for a week or two and revisit this after I have internalized it better!

SeanTAllen:
@dusty-phillips I don't have time to pick this up right now, but I do think that you have highlighted the need for a "defaults" section "somewhere". Probably around where you placed it but maybe not. While you are exploring, please do avail yourself of the beginners help stream in Zulip. It's a great place to get additional answers. There's definitely parts of your understanding that aren't correct, but that isn't problematic for where you are in the learning curve. I'd be happy to assist with the finer points on Zulip if you ever want to take me up on that.

redvers:
I confess that as a pony developer I've always written my refcaps explicitly because I found defaults so difficult to internalize.

This is a massively useful idea - thank you so much for writing it.

dusty-phillips:
Clearing PRs I don't intend to maintain out of my dashboard.

redvers:
@dusty-phillips - thank you so much for the effort you've put in thus far.  I'm absolutely behind what you were trying to do here and I'm going to see if I can find some way to make it fit.  It's the kind of thing I wanted to put on page 2 on a pony cheat-sheet because - like you said in your docs it's the kind of thing that new pony programmers are going to want available to reference.

So - thanks again!

PR #507, opened 2022-03-20
Generic 'get' function should have default refcap

djavorszky:
Having `fun ref get(): this->A => _c` results in compile error:

> Error:
/home/djavorszky/dev/helloworld/main.pony:7:18: can only use 'this' for a viewpoint in a box function
  fun ref get(): this->A => _c

Works with `fun get(): this->A => _c`


SeanTAllen:
Thanks

PR #508, opened 2022-03-21
Document return-type polymorphism in FFI

ergl:
This documents the new feature to be introduced in https://github.com/ponylang/ponyc/pull/4060, so this shouldn't be merged until that becomes available in a release.

### DIFF
--- a/docs/c-ffi/calling-c.md
+++ b/docs/c-ffi/calling-c.md
@@ -24,7 +24,7 @@ class val FilePath
       end
 ```
 
-FFI functions have the @ symbol before its name, and FFI signatures are declared using the `use` command. The types specified here are considered authoritative, and any FFI calls that differ are considered to be an error.
+FFI functions have the @ symbol before its name, and FFI signatures are declared using the `use` command. The types specified here are considered authoritative, and any FFI calls that use different types will result in a compile error.
 
 The use @ command can take a condition just like other `use` commands. This is useful in this case, since the `_mkdir` function only exists in Windows.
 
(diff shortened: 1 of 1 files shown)


ergl:
@redvers Let me know if this new section would have helped you with your issues. It doesn't cover generics since the interplay between that and FFI is a bit complicated from the type checking perspective.

docs/c-ffi/calling-c.md
```diff
@@ -24,7 +24,7 @@ class val FilePath
       end
 ```
 
-FFI functions have the @ symbol before its name, and FFI signatures are declared using the `use` command. The types specified here are considered authoritative, and any FFI calls that differ are considered to be an error.
+FFI functions have the @ symbol before its name, and FFI signatures are declared using the `use` command. The types specified here are considered authoritative, and any FFI calls that use different types will result in a compile error.
```
SeanTAllen, line 27:
```suggestion
FFI functions have the @ symbol before its name, and FFI signatures are declared using the `use` command. The types specified here are considered authoritative, and any FFI calls that use different parameter types will result in a compile error.
```

docs/c-ffi/calling-c.md
```diff
@@ -153,6 +153,73 @@ env.out.print(size.height.string())
...
+```pony
+// Compiler error: couldn't find 'x' in 'Pointer'
+let point_x = @list_pop(list_of_points)
+point.x
+
+// Compiler error: wanted Pointer[U8 val] ref^, got Pointer[None val] ref
+let head = String.from_cstring(@list_pop(list_of_strings))
+```
+
+We can fix this problem by adding an explicit return type parameter when calling `list_pop`:
```
SeanTAllen, line 207:
```suggestion
We can fix this problem by adding an explicit return type when calling `list_pop`:
```

docs/c-ffi/calling-c.md
```diff
@@ -153,6 +153,73 @@ env.out.print(size.height.string())
...
+let x_coord = point.x
+
+// OK
+let pointer = @list_pop[Pointer[U8]](list_of_strings)
+let data = String.from_cstring(pointer)
+```
+
+Note that the declaration for `list_pop` is still needed: if we don't add an explicit return type when calling `list_pop`, the default type will be the return type of the declaration.
+
+You'll notice that this feature enables you to cast the return type of a C function to any other type, **making this feature potentially dangerous if you cast to the wrong type**. As we have noted before, when it comes to FFI, the Pony compiler trusts that you know what you are doing, so you MUST get the return type right.
```
SeanTAllen, line 221: So there's nothing about this that is any more scary than any other C interface but it feels like it is making the override sound more scary. I think this paragraph can be removed. If we don't have "C-ffi is a big foot gun" elsewhere in the tutorial we should add it as an "in general" topic for FFI rather than this note here that makes the override sound more dangerous than it is.

ergl: We do have a big warning at the beginning of the chapter, but I thought it was worth repeating this since this allowed unconstrained type casts. 


That said, after reading @jemc's comment about adding ABI size checks, I would feel more comfortable with removing this paragraph.

PR #511, opened 2022-04-08
Brought examples up to date with current Pony syntax and package requirements.

redvers:
* appendices/examples.md - UnitTest.apply() returns None, not TestResult.
  expect_eq method is now assert_eq
* expressions/control-structures.md - typo in the variable name.
* gotchas/garbage_collection.md - Missing => in constructor line,
  missing use "collections" needed for Range.
* pattern-matching/match.md - Mandatory constructor for Main added.

### DIFF
--- a/docs/appendices/examples.md
+++ b/docs/appendices/examples.md
@@ -129,8 +129,8 @@ class iso _TestAddition is UnitTest
   """
   fun name(): String => "u32/add"
 
-  fun apply(h: TestHelper): TestResult =>
-    h.expect_eq[U32](2 + 2, 4)
+  fun apply(h: TestHelper) =>
+    h.assert_eq[U32](2 + 2, 4)
 ```
 
 Some assertions you can make with `TestHelper` are
--- a/docs/expressions/control-structures.md
+++ b/docs/expressions/control-structures.md
@@ -140,7 +140,7 @@ actor Main
       else
         "no names!"
       end
-    env.out.print("x is " + s)
+    env.out.print("x is " + x)
 ```
 
 And finally, here the value of __x__ is "no names!"
--- a/docs/gotchas/garbage-collection.md
+++ b/docs/gotchas/garbage-collection.md
@@ -15,8 +15,10 @@ Garbage collection is never attempted on any actor while it is executing a behav
 Here's a typical "I'm learning Pony" program:
 
 ```pony
+use "collections"
+
 actor Main
-  new create(env: Env)
+  new create(env: Env) =>
     for i in Range(1, 2_000_000) do
       ... something that uses up heap ...
     end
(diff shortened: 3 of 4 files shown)


redvers:
I also noticed that the list of additional functions for TestHelper on the examples page contained a few methods that no longer exist and those that do exist as fun tag.

The pony_test documentation doesn't have the functions as fun tag… and it looks like the reason why that is is that TestHelper used to be an actor.

I just wanted to make sure that there wasn't any viewpoint adaptation or some other reason why they would be fun tag in the tutorial but not in the docs before I modified that too.

redvers:
> @redvers Will you list a couple of the functions that: a) no longer exist, and b) are now fun tag? I will take a look, but want to ensure I am working from the same list as you.

These from https://tutorial.ponylang.io/appendices/examples.html#how-to-write-tests
image

No longer exist:
* be fail()
* be assert_failed(msg: String)
* fun tag expect_true
* fun tag expect_false
* fun tag expect_error
* fun tag expect_is

Wrong refcap:
* fun log

Wrong refcap & args:
* fun assert_true
* fun assert_false
* fun assert_error
* fun assert_is
* fun assert_eq

rhagenson:
I cannot think of any reason they should be listed differently between the tutorial and documentation with regard to viewpoint adaptation or otherwise. (I also do not like the way that list in the tutorial is right now with regard to "Some assertions you can make with TestHelper are" being followed by a list with non-assertion methods.)

I need to review the rules around `fun tag` for my own purposes, but for moving this PR forward I cannot find any reason the Tutorial should not match stdlib documentation.

SeanTAllen:
This is why in thought examples shouldn't be in the tutorial

rhagenson:
I does feel odd to have examples. I think right now we are in the unfortunate space of needing examples to be somewhere and the tutorial being the most logical place for them. I have alluded to it on various calls, but one of my plans once I get RFC 200 (and other work I have already begun) completed, I want to work on an example-driven complement to the Tutorial.

SeanTAllen:
I think all examples should be moved to the ponyc repo where we already have some and ci runs to make sure they work.

SeanTAllen:
Or they should be part of standard library documentation. I think red has demonstrated pretty well that this section should be removed. It's at best duplicating information that should be elsewhere. These aren't particularly good examples either. They lack the explanation we have in both patterns and stdlib documentation and even the ponyc examples. 

I'd rather delete this section and if there is anything here that people think is something that could have value, open issues for the appropriate creation.

rhagenson:
I think having these examples in `ponyc/examples/` makes sense. I also think removing them as part of this PR is not fitting.

redvers:
I agree.  I would agree that that removal should be separate.

Are there any other changes you need in this PR before approval?

rhagenson:
I don't know of anything remaining to be done for this PR, however before this is merged will you change the title to reflect what these "small number of edits" are intended to do? The primary purpose is to make each modified block executable, yes?

PR #512, opened 2022-05-19
Improve phrasing for the serialisation section

ergl:


docs/appendices/serialisation.md
```diff
@@ -59,7 +59,7 @@ actor Main
 
 There are several things to keep in mind when using Pony's serialisation system:
 
-* Serialised objects will currently only work when passed between two running instances of the same Pony executable. You cannot pass objects between different Pony programs, nor can you pass them between different versions of the same program.
+* Serialised objects will currently only work when passed between two running instances of the same Pony executable. Passing objects between different Pony programs or between different versions of the same program may lead to a crash or unsafe memory manipulation. You might use the `Serialise.signature` function to determine that sharing two different Pony programs is safe.
```
SeanTAllen, line 62: While this is technically true. It is highly unlikely that will ever be true. I think the "less true" version is "more true in practice".

docs/appendices/serialisation.md
```diff
@@ -86,11 +86,11 @@ The custom deserialisation method is expected to modify the values of the object
...
 
 #### Classes With Multiple `Pointer` Fields
 
-If a class has more than one `Pointer` field then all of those fields must be handled by the custom serialisation and deserialisation methods for that class; there are not methods for each field. For example, if a class has three `Pointer` fields then the `_serialise_space()` method must return the total number of bytes required to serialise the objects from all three fields.
+If a class has more than one `Pointer` field then all of those fields must be handled by the custom serialisation and deserialisation methods for that class; there are not methods for each field. For example, if a class has three `Pointer` fields, then the `_serialise_space()` method must return the total number of bytes required to serialise the objects from all three fields.
```
SeanTAllen, line 93: what changed here?

docs/appendices/serialisation.md
```diff
@@ -86,11 +86,11 @@ The custom deserialisation method is expected to modify the values of the object
...
+The programmer must write their custom serialisation and deserialisation code in such a way that it is aware of how many bytes are available in the byte arrays that are passed to the methods. If the objects are always of a fixed size then the functions can read and write that many bytes to the buffer. However, if the objects are of varying sizes (for example, if the object was a string), then the serialized representation must include information that the deserialisation code can use to ensure that it does not read beyond the end of the memory occupied by the object. The custom serialisation system does not provide a mechanism for doing this, so it is up to the program to choose a mechanism and implement it. In the case of a string, the serialisation format could consist of a 4-byte header that encodes the length of the string, followed by a string of the specified length. These additional four bytes must be included in the value returned by `_serialise_space()`. The deserialisation function would then start by reading the first four bytes of the array to obtain the size of the string and then read only that many bytes from the array.
```
SeanTAllen, line 89: what changed here?

docs/appendices/serialisation.md
ergl: I added a comma after "if a class has three `Pointer` fields", which I think improves the flow of reading

docs/appendices/serialisation.md
ergl: Do you mean the "crash" part, or the part about using the signature to know if it is safe?

ergl: Also, just noticed that there's a missing "data" in the last sentence

docs/appendices/serialisation.md
SeanTAllen: This one looks good. Nice.

docs/appendices/serialisation.md
SeanTAllen: I'm not sure that is gramatically correct. I'm not sure it isn't

ergl: I have reverted this, since at least we know that the original version is correct

docs/appendices/serialisation.md
SeanTAllen: Technically you might be able to pass objects between different versions of a program but it is highly unlikely. In both those instances, the chance that the type ids are the same is really low. So that "may lead to" is more like "will almost assuredly".

So while your rewrite is technically more correct, I think it is less practically correct. I dont think the rewrite properly conveys the danger and makes it sound easier to accomplish than it is. Current serialization is not intended to be used in the way your rewrite suggests. It was intended in the "only like this" that the original version conveys.

SeanTAllen:
@ergl the linting error should be resolved in some fashion before merging.

docs/appendices/serialisation.md
ergl: Okay, I've reverted to the original version, but added a note about using the `signature` function.

PR #513, opened 2022-05-19
Set lint for unordered list indentation to 4 spaces

ergl:
The default value is 2, but mkdocs needs at least 4 spaces in order to correctly
render nested lists. See https://github.com/mkdocs/mkdocs/issues/545 for more
details.


SeanTAllen:
I am approving this assuming all linting passes with the change, if not, then the failures would need to be fixed as part of this PR.

SeanTAllen (approved):

PR #518, opened 2023-09-05
Add code copy button to codeblocks

SeanTAllen:


SeanTAllen:
This sets them up on all codeblocks. If desired, we can add to only the blocks we want on OR better if we dont want on all, we can identify which blocks we dont want on and turn off on those. See: https://squidfunk.github.io/mkdocs-material/reference/code-blocks/?h=code+copy+button#code-copy-button

PR #522, opened 2023-12-10
Build the tutorial nightly to get alerts if it fails

SeanTAllen:
Currently, a person opening a PR can be surprised by failures unrelated
to their changes. In particular, to URLs that are external to the site
no longer working and failing validation.

Worse, this means that users of the tutorial are experiencing broken
URLs that won't be discovered until someone does an update for the
tutorial and the build fails.

This commit adds a nightly job that tries to build the site and if it
fails, alerts to Zulip so we can be made aware of issues in a timely
fashion and can endeavor to fix them as soon as possible.

The workflow can also be triggered manually if need be.

### DIFF
--- a/.github/workflows/verify-site-builds.yml
+++ b/.github/workflows/verify-site-builds.yml
@@ -1,0 +1,31 @@
+name: Verify site builds
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '0 2 * * *'
+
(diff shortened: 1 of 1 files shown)


rhagenson (approved):
Looks good to me!

chalcolith (approved):

jemc (approved):

PR #525, opened 2023-12-29
calling-c: elaborate on the x_dpy example

greenfork:
The following code is not valid Pony code

    let x_dpy = @XOpenDisplay(Pointer[U8])

### DIFF
--- a/docs/c-ffi/calling-c.md
+++ b/docs/c-ffi/calling-c.md
@@ -80,7 +80,8 @@ use @eglGetDisplay[Pointer[_EGLDisplayHandle]](disp: Pointer[_XDisplayHandle])
 primitive _XDisplayHandle
 primitive _EGLDisplayHandle
 
-let x_dpy = @XOpenDisplay(Pointer[U8])
+let x_dpy_name: String = // ...
+let x_dpy = @XOpenDisplay(x_dpy_name.cstring())
 if x_dpy.is_null() then
   env.out.print("XOpenDisplay failed")
 end


docs/c-ffi/calling-c.md
```diff
@@ -80,7 +80,8 @@ use @eglGetDisplay[Pointer[_EGLDisplayHandle]](disp: Pointer[_XDisplayHandle])
 primitive _XDisplayHandle
 primitive _EGLDisplayHandle
 
-let x_dpy = @XOpenDisplay(Pointer[U8])
+let x_dpy_name: String = // ...
```
SeanTAllen, line 83: So the code here already "doesn't compile" as it isn't valid. I don't think this change makes things better.

I think that this could be slightly better if the string is given a realistic display name.
This could be made "much better" by taking all the C calling code and make it "real" by putting the "FFI code examples" in code that would at least compile.

SeanTAllen:
@ponylang/committer I think the examples in this section are kind of a mess. They aren't valid code and leave alot for someone to guess how they are correct. I don't think that this change is helpful but I think it does highlight for me that the exmaples in this section are problematic.

I'd prefer to kick out an issue to fix the "this is basically Pony pseudo code" issues in this section.

If other committers feel this change is an improvement, we can both merge this and kick out an issue.

SeanTAllen:
Closed in favor of https://github.com/ponylang/pony-tutorial/issues/526

PR #527, opened 2024-01-13
Clarify that `struct` is always a pointer

SeanTAllen:
It wasn't clear that a struct as a return value from C should be read as "pointer to struct".

### DIFF
--- a/docs/c-ffi/calling-c.md
+++ b/docs/c-ffi/calling-c.md
@@ -153,7 +153,7 @@ env.out.print(size.height.string())
 
 A `NullablePointer` type can only be used with `structs`, and is only intended for output parameters (like in the example above) or for return types from C. You don't need to use a `NullablePointer` if you are only passing a `struct` as a regular input parameter.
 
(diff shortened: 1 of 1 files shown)


ergl (approved):

docs/c-ffi/calling-c.md
```diff
@@ -153,6 +153,20 @@ env.out.print(size.height.string())
 
 A `NullablePointer` type can only be used with `structs`, and is only intended for output parameters (like in the example above) or for return types from C. You don't need to use a `NullablePointer` if you are only passing a `struct` as a regular input parameter.
 
+If you are using a C function that returns a struct, remember, that you need to return a pointer to the struct. The following in Pony should be read as **returns a pointer to struct `Rect`**:
```
jemc, line 156:
```suggestion
If you are using a C function that returns a struct, remember, that the C function needs to return a pointer to the struct. The following in Pony should be read as **returns a pointer to struct `Rect`**:
```

PR #532, opened 2024-02-09
Improve example of `with` block with multiple objects

i-ky:
Emphasize that all disposable objects listed in the first line of `with` block can be used within the block.


SeanTAllen:
Thanks @i-ky. Nice small improvement that can go a long way towards eliminating a possible source of confusion.

PR #539, opened 2024-03-26
Move "Match Expressions" and "As Operator" (in that order) into the "Expressions" section, between "Control Structures" and "Methods"

### LINKED ISSUE #536: Tutorial chapter order
When reading the tutorial, even though, I did it in the order, the chapters were provided, it sometimes felt, like I actually read them out of order. Here are a few suggestions that would—at least for me—would feel more intuitive:
    – Explain [literals](https://tutorial.ponylang.io/expressions/literals) _before_ [classes](https://tutorial.ponylang.io/types/classes) 
    – Move [trust boundary](https://tutorial.ponylang.io/object-capabilities/trust-boundary) to [C-FFI](https://tutorial.ponylang.io/c-ffi/)
    – Move [matching](https://tutorial.ponylang.io/pattern-matching/match) to [operators](https://tutorial.ponylang.io/expressions/ops)

I have a hunch, with that, the number of "we'll explain later" and "we partially already explained this earlier" can be reduced

This might be related to #14

shaedrich:
Fix #536

As discussed in https://github.com/ponylang/pony-tutorial/issues/536#issuecomment-2021173572


SeanTAllen:
This should include redirects from old location to new in netlify.toml

SeanTAllen:
There's errors on the build side with netlify:

https://app.netlify.com/sites/pony-tutorial/deploys/66032268a3fe2b0009664180

if some of those are internal links than rather than recovering with the redirects (which we want), it would be ideal to fix the internal links as well to go to the correct location rather than following a redirect.

shaedrich:
Should work now: https://deploy-preview-539--pony-tutorial.netlify.app/expressions/pattern-matching/

SeanTAllen:
The pattern matching subdirectory should be removed and the overview deleted. 

That was the intent. Sorry that wasn't communicated more clearly.

shaedrich:
Yeah, sorry, I wasn't sure. But no problem, it's fixed now.

netlify.toml
```diff
@@ -42,3 +42,9 @@ ID = "Your_Site_ID"
   to = "/reference-capabilities/:splat"
   status = 301
   force = true
+
+[[redirects]]
+  from = "/pattern-matching/*"
+  to = "/expressions/:splat"
+  status = 301
+  force = true
```
SeanTAllen, line 50: can you update this to be the specific urls to redirect? if we ever change these again, it will be to update again (potentially) and can you send the old overview/index to the match.md?

SeanTAllen:
One more small change. Sorry @shaedrich. Thanks for your work on this.

netlify.toml
shaedrich: Sure, not problem 👍🏻

shaedrich:
No trouble at all. You're welcome.

SeanTAllen:
Thanks @shaedrich

shaedrich:
My pleasure, @SeanTAllen

PR #548, opened 2024-05-24
Meta Tags, OpenGraph and Twitter Cards

### LINKED ISSUE #546: Meta Tags, OpenGraph and Twitter Cards
Wouldn't it be nice, if the ponylang sites had OpenGraph and such? This could be done via the [meta plugin](https://mrkeo.github.io/reference/meta-tags/) ([squidfunk docs](https://squidfunk.github.io/mkdocs-material/plugins/meta/), [Python Markdown docs](https://python-markdown.github.io/extensions/meta_data/)) and by [extending the theme](https://mrkeo.github.io/customization/#extending-the-theme).

shaedrich:
Fixes #546

### DIFF
--- a/.github/linters/.markdown-lint.yml
+++ b/.github/linters/.markdown-lint.yml
@@ -1,8 +1,5 @@
 {
   "MD007": { "indent": 4 },
   "MD013": false,
-  "MD026": false,
-  "MD025": {
-    front_matter_title: ""
-  }
+  "MD026": false
 }
--- a/.github/linters/.markdown-lint.yml
+++ b/.github/linters/.markdown-lint.yml
@@ -1,5 +1,8 @@
 {
   "MD007": { "indent": 4 },
   "MD013": false,
-  "MD026": false
+  "MD026": false,
+  "MD025": {
+    front_matter_title: ""
+  }
 }
--- a/docs/getting-started/index.md
+++ b/docs/getting-started/index.md
@@ -1,5 +1,4 @@
 ---
-title: Getting Started – Overview
 description: This chapter will get you up and running with Pony from installing the compiler to running your first program.
 ---
 # Overview
--- a/docs/getting-started/index.md
+++ b/docs/getting-started/index.md
@@ -1,3 +1,7 @@
+---
+title: Getting Started – Overview
+description: This chapter will get you up and running with Pony from installing the compiler to running your first program.
+---
 # Overview
 
 This chapter will get you up and running with Pony from installing the compiler to running your first program.
--- a/layouts/default.yml
+++ b/layouts/default.yml
@@ -1,0 +1,22 @@
+definitions:
+  - &page_title_with_site_name >-
+    {% set title = config.site_name %}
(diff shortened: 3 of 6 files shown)


shaedrich:
Result: 😍 
grafik

SeanTAllen:
One of my favorite things about when we moved from the Hugo theme I did that supported meta tags and twitter cards was that we didn't have to try to come up with descriptions for every page. It was a slog in some cases. 

We will discuss at sync and perhaps other maintainers will see the value. I personally have been down the road of this already with our old setup and I didn't see the value in it. It was just extra work, extra stuff to maintain so I'm not in favor, but others might be. 

Will update further after our next sync call.

Thanks for opening this PR Sebastian. Your commit to improving the tutorial has been impressive and inspiring to watch as the updates flow into my inbox.

shaedrich:
> One of my favorite things about when we moved from the Hugo theme I did that supported meta tags and twitter cards was that we didn't have to try to come up with descriptions for every page. It was a slog in some cases.

You unfortunaly see it from a slightly "wrong" angle. I totally see, that you don't want to write these descriptions. But I wouldn't call a SSG engine "better" when it is lacking a fairly standard feature, just because I personally don't like the work that goes into it and I, unlike many other people, don't use that feature.

Sorry if I'm out of line here.

shaedrich:
> Can the title come from the H1 in the page? Having in metadata, it could get out of sync.

It _could_. However, as you see, this works fine on page but less so more often than not taken out of context as it will be when sent as a link.

SeanTAllen:
> It _could_. However, as you see, this works fine on page but less so more often than not taken out of context as it will be when sent as a link.

Can it come from the existing breadcrumb stuff in mkdocs?

jemc:
I'm in agreement with @SeanTAllen on this point.

I admit that having page-specific information in the Card can be useful for someone glancing quickly card but it's a fairly small amount of value and I don't think it outweighs the maintenance effort being added.

Here are my concrete set of concerns:
- the problem of drift between in the title in the metadata vs the title in the H1
- the problem of drift between chapter title in the metadata vs the actual title of the containing chapter
- the problem where all these meta fields are hidden during normal browsing, and someone may not notice that they need to update them when only looking at the static site in the netlify preview.
- the description copy, being hidden may not get as much attention as it perhaps should during maintenance

Some alternative ideas discussed in sync:
- try to pull the title info from the H1 and the chapter title, instead of duplicating that info
- use a generic description for the entire tutorial, or perhaps a per-chapter description as a compromise.

shaedrich:
> > It _could_. However, as you see, this works fine on page but less so more often than not taken out of context as it will be when sent as a link.
> 
> Can it come from the existing breadcrumb stuff in mkdocs?

Can you elaborate on using this breadcrumb stuff? I'm not sure how I would tackle this, using pre-existing data

shaedrich:
See https://github.com/ponylang/pony-tutorial/issues/546#issuecomment-2189685936

Issue #340, opened 2019-01-05
Add ability to run code examples in the playground


SeanTAllen:
We had this sort of on the old tutorial but that didn't get migrated over.

SeanTAllen:
I went looking through the mkdocs material for this functionality and didn't see it. I'm going to reach out to the author and see if I missed the functionality, otherwise, I'll put in a feature request.

SeanTAllen:
So this is the report:

We can do some custom css and javascript. That's not really my thing, but someone could take this on.

https://github.com/squidfunk/mkdocs-material/discussions/3563

shaedrich:
There doesn't seem to be a native option to [add any other button than "copy" and "select"](https://squidfunk.github.io/mkdocs-material/reference/code-blocks/#configuration) with the latter being sponsors-only:
![grafik](https://github.com/ponylang/pony-tutorial/assets/11225821/3efbf323-4371-4fab-93e8-6ac094e36681)

> I went looking through the mkdocs material for this functionality and didn't see it. I'm going to reach out to the author and see if I missed the functionality, otherwise, I'll put in a feature request.

Others have also wished for this functionality to become more generalized: https://github.com/squidfunk/mkdocs-material/discussions/4005 The only solution given there is to fork the theme.

However, one _could_ add custom JavaScript to add this behavior: https://squidfunk.github.io/mkdocs-material/customization/#additional-javascript

![grafik](https://github.com/ponylang/pony-tutorial/assets/11225821/e8eb9e53-52bf-4200-ad7f-2bfea514ca0a)

To achieve a quick and easy solution, we could add buttons to all code blocks manually

![grafik](https://github.com/ponylang/pony-tutorial/assets/11225821/f6311ffa-009c-4e60-98eb-38b54d5b24c0)

Something, that would be more work but would prevent us from duplicating the code (once inside the code block and once URL-encoded in the button), would be to move the code to separate files and [embed them](https://squidfunk.github.io/mkdocs-material/reference/code-blocks/#embedding-external-files)

This could also help in cases where the examples are longer than the URL is allowed to be by the browser (I don't know if we would run into that problem) 🤔

SeanTAllen:
We use the insiders build. So that is available as an option.

shaedrich:
Yeah, but that is only the "select" button, not the "play"/"execute" button

jemc:
[...]

PR #545, opened 2024-05-20
[2/?] Add ability to run code examples in the playground: Extend code samples to be runnable

shaedrich:
See #340

As per https://github.com/ponylang/pony-tutorial/pull/542#issuecomment-2113695598

# Stats
* **runnable:** 148
* **not runnable:** 157
* **erroring:** 6
* **open questions:** 3

# Files
## Getting Started
* [x] hello-world-main.pony &rarr; prints <blockquote>Hello, world!</blockquote>
    * [x] hello-world.md: `--8<-- "hello-world-main.pony` &rarr; runnable
    * [x] how-it-works.md: `--8<-- "hello-world-main.pony` &rarr; runnable
    * [x] how-it-works.md: `--8<-- "hello-world-main.pony::1` &rarr; not runnable
    * [x] how-it-works.md: `--8<-- "hello-world-main.pony:2:2` &rarr; not runnable
    * [x] how-it-works.md: `--8<-- "hello-world-main.pony:3:3` &rarr; not runnable

## Types
* [x] classes-swap-values-sugar.pony &rarr; prints <blockquote>a = 1, b = 2
a = 2, b = 1</blockquote>
    * [x] `--8<-- "classes-swap-values-sugar.pony`<ins>`:6:6`</ins>`"` &rarr; runnable
* [x] classes-swap-values.pony &rarr; prints <blockquote>a = 1, b = 2
a = 2, b = 1</blockquote>
    * [x] `--8<-- "classes-swap-values.pony`<ins>`:6:8`</ins>`"` &rarr; runnable
* [ ] ~~classes-wombat-constructor-invocation.pony~~ _(see classes-wombat.pony)_
* [x] classes-wombat-constructors.pony &rarr; prints <blockquote>Your default wombat's name is "Fantastibat"
Your hungry wombat's name is "Nomsbat"</blockquote>
    * [x] `--8<-- "classes-wombat.pony`<ins>`:8:18`</ins>`"` &rarr; runnable
* [x] classes-wombat.pony &rarr; prints <blockquote>Fantastibat has a hunger level of 5
Nomsbat has a hunger level of 12</blockquote>
    * [x] `--8<-- "classes-wombat.pony:`<del>`1:1`</del><ins>`9:9`</ins>`"` &rarr; not runnable
    * [x] `--8<-- "classes-wombat.pony:`<del>`1:3`</del><ins>`9:11`</ins>`"` &rarr; not runnable
[...]

PR #542, opened 2024-04-22
[3/?] Add ability to run code examples in the playground: Fetch tutorial snippet from URL by file name

shaedrich:
See https://github.com/ponylang/pony-tutorial/issues/340

> [!WARNING]
Will only work, after #544 has been merged. Works in combination with https://github.com/ponylang/pony-playground/pull/205. Contains changes from #544 until db1eb1183d0746535db719cd537ad8c00a9cc1d1

**Example:** https://deploy-preview-542--pony-tutorial.netlify.app/getting-started/hello-world#the-code

# Open questions
* ~~What to do with the non-pony files?~~
* What to do with partial examples that won't run on their own?
* > no Main actor found in package 'main'

# Files
## Getting Started
* [x] hello-world-main.pony

## Types
* [ ] classes-swap-values-sugar.pony
* [ ] classes-swap-values.pony
* [ ] classes-wombat-constructor-invocation.pony
* [ ] classes-wombat-constructors.pony
* [ ] classes-wombat.pony
* [ ] classes-zero-argument-constructors-invocation.pony
* [ ] classes-zero-argument-constructors.pony
* [ ] primitives-doors.pony
* [ ] actors-behaviors.pony
* [ ] actors-sequential.pony
* [ ] traits-and-interfaces-marker-methods.pony
* [ ] traits-and-interfaces-multiple-traits.pony
* [ ] traits-and-interfaces-nested-traits.pony
* [ ] traits-and-interfaces-nominal-and-structural-subtyping.pony
* [ ] traits-and-interfaces-nominal-subtyping-in-pony.pony
* [ ] traits-and-interfaces-nominal-subtyping.pony
* [ ] traits-and-interfaces-open-world-enumerations.pony
* [ ] traits-and-interfaces-open-world-interface.pony
* [ ] traits-and-interfaces-open-world-typing.pony
* [ ] traits-and-interfaces-private-methods.pony
* [ ] traits-and-interfaces-structural-subtyping-in-pony.pony
* [ ] traits-and-interfaces-structural-subtyping.pony
* [ ] traits-and-interfaces-trait.pony
* [ ] traits-and-interfaces-type-union.pony
* [ ] structs-constructors.pony
* [ ] structs-fields.pony
* [ ] type-aliases-complex-types-interface.pony
* [ ] type-aliases-complex-types-trait.pony
* [ ] type-aliases-enumerations-apply.pony
* [ ] type-aliases-enumerations-iteration.pony
* [ ] type-aliases-enumerations-namespace.pony
[...]

PR #550, opened 2024-05-31
[4/?] Add ability to run code examples in the playground: Get pony snippets tested with ponyc at CI time

### LINKED ISSUE #549: [4/?] Add ability to run code examples in the playground: Get pony snippets tested with ponyc at CI time
* Where do we put the expectations for `stdOut`, `stdErr`, and `exitCode`? Because if we don't do that, only snippets that print to `stdOut` could be tested 

| Chapter | Snippet | stdOut | stdErr | exitCode |
| ------- | ------- | ------ | :----: | :------: |
| Getting started &#9656; Hello world | hello-world-main.pony | Hello, world! | _n/a_ |    0     |

shaedrich:
Fix #549

See #340

### DIFF
--- a/.github/workflows/check-code-samples.sh
+++ b/.github/workflows/check-code-samples.sh
@@ -1,0 +1,65 @@
+cd ./code-samples/
+files=$(ls | wc -l)
+echo "Check $files files …"
+failedFiles=()
+i=0
+for file in *.pony; do # actors-sequential.pony
+    ((i++))
+    percentage=$(((i*100)/files))
+    echo -e "#$i Test $file … ($i/$files \u2192 $percentage %)"
+    contents=$(cat "$file") # code-samples/
+    #echo "contents: $contents"
+    regex='^"""\n(?<line>^(?<key>[a-z_]+):\s?(?<value>.*?)$\n)+"""' # TODO: Get frontmatter from AST
+    regex='^"""\n(.*)\n"""'
+    expectations=$(jq --arg file "${file}" ".[\"$file\"]" ../code-samples.json) # -r
+    #echo "Expectations for $file $expectations"
+    expectedStdout=$(echo "$expectations" | jq '.stdout')
+    expectedStderr=$(echo "$expectations" | jq '.stderr')
+    expectedExitcode=$(echo "$expectations" | jq '.exitcode')
+    #echo "stdout for $file: "
+    #echo "$contents" | grep -e $regex
+    if [ $(echo "$contents" | grep -qe $regex) ]; then # TODO: fix regex
+        frontmatter="${BASH_REMATCH[0]}"
+        readarray -d "\n" -t lines <<< frontmatter
+        echo "Expectations 1: ${lines[0]}"
+    fi
+    if [[ "$contents" =~ $regex ]]; then
+        frontmatter="${BASH_REMATCH[0]}"
+        readarray -d "\n" -t lines <<< frontmatter
+        echo "Expectations: ${lines[0]}"
+    fi
+    encoded=$(sed -E -e ':a;N;$!ba;s/\r\n|\r|\n/\\n/g' "$file") #code-samples/
+    encoded=${encoded//\"/\\\"}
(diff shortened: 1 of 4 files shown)


shaedrich:
fyi: The new job fails because currently only one code snippet has expectations set (all of them have to)

.github/workflows/check-code-samples.sh
```diff
@@ -0,0 +1,75 @@
...
+    fi
+    encoded=$(sed -E -e ':a;N;$!ba;s/\r\n|\r|\n/\\n/g' "$file") #code-samples/
+    encoded=${encoded//\"/\\\"}
+    encoded=${encoded//   /\\t}
+    json="{\"code\": \"$encoded\", \"separate_output\": true, \"color\": true, \"branch\": \"release\"}"
+    #echo "Send $json…"
+    response=$(curl -s --header "Content-Type: application/json" \
+        --request POST \
+        --data "$json" \
+        "https://playground.ponylang.io/evaluate.json")
```
jemc, line 46: Rather than calling out to the playground to run these, we should use `ponyc` directly here. We don't want to flood the playground when running these tests, and we don't want the uptime of the playground to affect these tests.

code-samples.json
```diff
@@ -0,0 +1,23 @@
...
+        "stdout": null,
+        "stderr": "interfaces can't have private methods, only traits can\n",
+        "exitcode": 1
+    },
+    "actors-behaviors.pony": {
+        "runnable": false
+    },
+    "appendices-examples-access-command-line-arguments.pony": {
+        "runnable": true,
+        "stdout": "This is printed first\nThis is printed last\n",
```
jemc, line 19: We shouldn't need to test `stdout` of running the examples - we just need to prove that an example successfully compiles (or fails to compile with a particular message in `stderr`). We don't need to run the examples at all - just compile them.

jemc (changes requested):

SeanTAllen:
For things that are "ci support scripts" our pattern is to put in .ci-scripts at the top level. If we have an external script like you currently have, it should go there, not in workflows.

That said, we have a pony program test runner in the ponyc repo that I think we should look at using instead of this script.

SeanTAllen:
When we add this, we should have a job that is triggered by the new release of a nightly image and new release of a release image to verify we dont have breakage. We have examples of this in a variety of library files. Once this basic on PR change is in, someone ping me and i'll add the "on update" workflows.

code-samples.json
shaedrich: What's the point of a snippet if it does something but we don't even know if it illustrates what the chapter talks about? It just happens to have a few buzzwords in it.

.github/workflows/check-code-samples.sh
shaedrich: Well, I implemented it in a way I'm familiar, since nobody replied to "I still need to figure out, how that might work 🤔 If there's someone else who would take this on, I wouldn't be mad 😅" in #340. I can try doing it via `ponyc` but how it will turn out is up to be shown.

shaedrich:
> When we add this, we should have a job that is triggered by the new release of a nightly image and new release of a release image to verify we dont have breakage. We have examples of this in a variety of library files. Once this basic on PR change is in, someone ping me and i'll add the "on update" workflows.

@SeanTAllen As explained above (https://github.com/ponylang/pony-tutorial/pull/550#discussion_r1626454322) to @jemc, the focus of this PR is not my strong suit and it would be great if someone else could tackle this

After all, I guess, [4/?] should start from scratch with a new PR as the implementation will vastly differ from what I tried.

jemc:
That's okay if you don't want to continue on this particular sub-task - we can try to find someone to pick up this work once your two remaining prior sub-tasks (i.e #545 , #542) are merged.

shaedrich:
#550 is not dependent on #542 so they can be worked on in parallel

shaedrich:
Okay, I guess, I managed to do an implementation with #551, which indeed has to wait for #545 to throw some less errors

PR #551, opened 2024-06-05
[4/?] Add ability to run code examples in the playground: Get pony snippets tested with ponyc at CI time

shaedrich:


shaedrich:
> invalid reference format
> _(From https://github.com/ponylang/pony-tutorial/actions/runs/9377723825/job/25819734031?pr=551)_

Does anyone know where that comes from?

Issue #558, opened 2025-10-01
may be a bug in object-literals lambdas section


glycerine:
https://tutorial.ponylang.io/expressions/object-literals.html?h=#lambdas

where it says

~~~
env.out.print(lambda("hello world"))
~~~

I note that this does not compile. I checked, and `lambda` is not itself a keyword; it is not listed here https://tutorial.ponylang.io/appendices/keywords.html

As a beginner, this is very confusing. 

I think it might be the result of a simple bug in the compilation of the code snippet that generates it.

I'm not sure what it should say, so I cannot suggest a fix off hand.

redvers:
It used to be a keyword - it was removed in [2021](https://github.com/ponylang/pony-tutorial/commit/2a1e0dd79915ad09032065f8d6489311178b4ab4)… but that's a coincidence.

It looks like when we moved all the example code into code-blocks that could be tested and was then referenced, a mistake crept in.

I'll get it PR'd.

But the correct line should be:

```pony
 {(s: String): String => "lambda: " + s }
```

PR #559, opened 2025-10-02
Fix line number reference in Lambda example

redvers:
Fixes https://github.com/ponylang/pony-tutorial/issues/558


SeanTAllen:
Thanks @redvers

PR #580, opened 2026-04-03
Document iftype

### LINKED ISSUE #393: Document `iftype`
It's been around for a long time as a "beta" sort of feature. It's required for some programs. At this point, I think we need to document it so that general users can access, use, and push our understanding of its shortcomings.

SeanTAllen:
Adds documentation for `iftype` to the control structures page, covering:

- Basic syntax and compile-time semantics
- Type narrowing within the `then` branch
- `elseif` chaining for multiple type checks
- Capability conditions (`ref`, `box`, etc.)
- Tuple conditions for checking multiple type parameters
- Limitations (left side must be a type parameter)
- Expression semantics (returns a value, implicit `else None`)

Includes 5 compilable code samples, updates to the keywords table and
symbol lookup cheat sheet, a cross-reference from the generics constraints
page, and a cspell dictionary addition.

### Parked review item

The basic example's `else` branch prints "woof" (implying a Cat-vs-Dog
dichotomy) rather than something type-neutral like "not a cat". The `else`
actually catches anything that isn't `Cat`. The prose explains this correctly
but the example could reinforce a wrong mental model. Low severity —
flagged during review as a pedagogical judgment call.

Closes #393

### DIFF
--- a/code-samples/control-structures-iftype-basic.pony
+++ b/code-samples/control-structures-iftype-basic.pony
@@ -11,5 +11,5 @@ actor Main
     iftype A <: Cat then
       env.out.print("meow")
     else
-      env.out.print("woof")
+      env.out.print("not a meow")
     end
--- a/code-samples/control-structures-iftype-basic.pony
+++ b/code-samples/control-structures-iftype-basic.pony
@@ -1,0 +1,15 @@
+trait Animal
+class Cat is Animal
+class Dog is Animal
+
+actor Main
+  new create(env: Env) =>
+    greet[Cat](env)
+    greet[Dog](env)
+
+  fun greet[A: Animal](env: Env) =>
+    iftype A <: Cat then
+      env.out.print("meow")
+    else
+      env.out.print("woof")
+    end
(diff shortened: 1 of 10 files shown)


code-samples/control-structures-iftype-basic.pony
```diff
@@ -0,0 +1,15 @@
...
+actor Main
+  new create(env: Env) =>
+    greet[Cat](env)
+    greet[Dog](env)
+
+  fun greet[A: Animal](env: Env) =>
+    iftype A <: Cat then
+      env.out.print("meow")
+    else
+      env.out.print("woof")
```
SeanTAllen:
```suggestion
      env.out.print("not a meow")
```

docs/expressions/control-structures.md
```diff
@@ -193,3 +193,71 @@ Suppose we're trying to create something and we want to keep trying until it's g
...
+
+__How is iftype different from match?__ A `match` expression dispatches on the runtime type of a value. `iftype` operates on type parameters as they are known at compile time. If you call a generic function `foo[Animal](Cat)`, the type parameter is `Animal`, not `Cat` — so `iftype A <: Cat` would be false, even though the runtime value is a `Cat`.
+
+Here's a simple example with a generic function that behaves differently depending on whether the type parameter is a specific class:
+
+```pony
+--8<-- "control-structures-iftype-basic.pony"
+```
+
+This prints "meow" and then "woof". When `greet` is called with `Cat`, the compiler sees that `Cat <: Cat` is true and takes the `then` branch. When called with `Dog`, `Dog <: Cat` is false, so it takes the `else` branch.
```
SeanTAllen:
```suggestion
This prints "meow" and then "not a meow". When `greet` is called with `Cat`, the compiler sees that `Cat <: Cat` is true and takes the `then` branch. When called with `Dog`, `Dog <: Cat` is false, so it takes the `else` branch.
```

jemc (approved):

